PR #6, opened 2024-08-29
Config paths

Surxe:
Config added and called instead of absolute paths.

Should readme info be moved to a section in root readme?

Tentatively changed 'python3' to 'python' until a more permanent solution is found


src/parser/decompile.sh
```diff
@@ -1,16 +1,18 @@
...
 
 mkdir -p "decompiled-data\localizations"
 cp "$DEADLOCK_LOCATION\game\citadel\resource\localization\citadel_gc\citadel_gc_english.txt" "decompiled-data\localizations\citadel_gc_english.txt"
-python3 localization_parser.py "decompiled-data\localizations\citadel_gc_english.txt" "decompiled-data\localizations\citadel_gc_english.json"
+python localization_parser.py "decompiled-data\localizations\citadel_gc_english.txt" "decompiled-data\localizations\citadel_gc_english.json"
 
 cp "$DEADLOCK_LOCATION\game\citadel\resource\localization\citadel_mods\citadel_mods_english.txt" "decompiled-data\localizations\citadel_mods_english.txt"
-python3 localization_parser.py "decompiled-data\localizations\citadel_mods_english.txt" "decompiled-data\localizations\citadel_mods_english.json"
\ No newline at end of file
+python localization_parser.py "decompiled-data\localizations\citadel_mods_english.txt" "decompiled-data\localizations\citadel_mods_english.json"
```
hassayag, line 18: We want to keep the use of `python3`, as linux/macos use `python` to usually mean python v2, so it will break this.

For your particular situation you can setup a new alias on your system.
1. Create `~/.bash_profile` if one doesn't already exist
2. Add `alias python3=python`
3. Open a fresh terminal

src/config/config_manager.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def get_path(key):
+    """Get a specific path from the config."""
+    config = load_config()
+    return config.get(key, '')
+
+
+if __name__ == '__main__':
+    import argparse
```
hassayag, line 56: imports should go at the top of the file

src/config/config_manager.py
```diff
@@ -0,0 +1,67 @@
...
+    parser = argparse.ArgumentParser(description='Manage user-specific paths.')
+    parser.add_argument('--set', nargs=2, metavar=('key', 'path'), help='Set a path.')
+    parser.add_argument('--get', metavar='key', help='Get a path.')
+
+    args = parser.parse_args()
+
+    if args.set:
+        set_path(args.set[0], args.set[1])
+    elif args.get:
+        print(get_path(args.get))
```
hassayag, line 67: This is really cool, but there is a simpler way of doing it that uses env vars, which allows for the following:
- Simple importing into the bash scripts
- Easier integration with prod envs and github actions

To do that, create a `.env.local` file in the root directory to look like
```
DEADLOCK_LOCATION="C:\Program Files (x86)\Steam\steamapps\common\Deadlock"
DECOMPILER_LOCATION=~/Downloads/Decompiler-windows-x64
```
then in any bash script you want to use that in, add  `. .env.local` at the top, which will load those variables, and can be referenced like `$DEADLOCK_LOCATION`.

Then for the setup part, we would have an `.env.example` to outline what `.env.local` looks like, as `.env.local` is gitignored

hassayag (changes requested):
Great idea to get this more friendly to new devs. I've pointed you in the direction environment variables, as that is an existing tool ready for the taking

Surxe:
Added the requested modifications.

`.env.local` is now used instead, and can be utilized by python functions in the new `src/utils/config_manager.py`

`python` changed back to `python3`

hassayag (changes requested):
Looks good, all that's left is to rework the python config file. Also do you know how I can check out your branch locally? I've never used the shared workspace, usually I'd just create a working branch

hassayag (approved):

PR #8, opened 2024-09-02
Infobox writer

Surxe:
Initial version of an infobox writer. Currently just for hero infoboxes, much room for expansion.


hassayag (approved):
Very slick tool for maintaining the infoboxes 👌

PR #15, opened 2024-09-10
Add OS check for decompile step and replace deprecated method

0xJas0n:
@hassayag i can't test this on Windows right now can you check if the path to the decompiler still works?

### DIFF
--- a/src/parser/decompiler/decompile.sh
+++ b/src/parser/decompiler/decompile.sh
@@ -8,13 +8,25 @@ FILES=("scripts/heroes" "scripts/abilities" "scripts/generic_data" "scripts/misc
 
 cp "$DEADLOCK_PATH/game/citadel/steam.inf" "decompiled-data/version.txt"
 
-# Loop through files and run Decompiler.exe for each
+ # Detect the operating system
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+      # MacOS
+      DECOMPILER_EXEC="Decompiler"  # Mac executable without .exe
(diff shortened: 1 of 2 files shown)


src/parser/decompiler/decompile.sh
```diff
@@ -8,13 +8,25 @@ FILES=("scripts/heroes" "scripts/abilities" "scripts/generic_data" "scripts/misc
...
+# Loop through files and run Decompiler for each
 for FILE in "${FILES[@]}"; do
   INPUT_PATH="$DEADLOCK_PATH/game/citadel/pak01_dir.vpk"
   VPK_FILEPATH="${FILE}.vdata_c"
-  
-  # Run the decompiler
-  "$DECOMPILER_PATH/Decompiler.exe" -i "$INPUT_PATH" --output "$OUTPUT_DIR" --vpk_filepath "$VPK_FILEPATH" -d
+
+# Run the decompiler
+"$DECOMPILER_PATH/$DECOMPILER_EXEC" -i "$INPUT_PATH" --output "$OUTPUT_DIR" --vpk_filepath "$VPK_FILEPATH" -d
```
hassayag, line 29: Could we instead just have `DECOMPILER_PATH` variable include the file name on the end? That means no need for any fancy OS detection

0xJas0n: Good call 👍. Ill probably get to it tomorrow

hassayag:
This PR is no longer needed, but let me know if you think otherwise @0xJas0n

PR #18, opened 2024-09-15
Feature/weapon stats

Surxe:
Parsed more weapon stats


src/parser/parsers/heroes.py
```diff
@@ -104,11 +104,60 @@ def _parse_hero_weapon(self, hero_value):
...
+                raise Exception('Bullet speed curve is not linear')
+
+        # Confirm its constant
+        last_y = spline[0]['y']
+        for point in spline:
+            if point['y'] != last_y:
+                raise Exception('Bullet speed curve is not constant')
+
+        # If constant, return the y
+        return last_y / ENGINE_UNITS_PER_METER
```
hassayag, line 159: I think we should round the bullet speed, as its been calculated empirically. So the many decimal places is a little misleading as to what our precision is

hassayag, line 159: I think we should round the bullet speed, as its been calculated empirically. So the many decimal places is a little misleading as to what our precision is

src/parser/parsers/heroes.py
```diff
@@ -104,11 +104,60 @@ def _parse_hero_weapon(self, hero_value):
...
+            'FalloffEndScale': weapon_prim['m_flDamageFalloffEndScale'],
+            'FalloffBias': weapon_prim['m_flDamageFalloffBias'],
+            'BulletGravityScale': weapon_prim['m_flBulletGravityScale'],
+            'BulletRadius': weapon_prim['m_flBulletRadius'] / ENGINE_UNITS_PER_METER,
         }
 
         weapon_stats['DPS'] = weapon_stats['BulletDamage'] * weapon_stats['RoundsPerSecond']
         return weapon_stats
 
+    def calc_bullet_velocity(self, spline):
```
hassayag, line 123: This is a private function, so add an underscore before it

hassayag (changes requested):
Really nice attention to future proofing the spline calculations, as who knows what could change

src/parser/parsers/heroes.py
```diff
@@ -202,3 +164,41 @@ def _map_attr_names(self, data, func):
...
+            raise Exception('Bullet speed curve is not linear')
+
+    # Confirm its constant
+    last_y = spline[0]['y']
+    for point in spline:
+        if point['y'] != last_y:
+            raise Exception('Bullet speed curve is not constant')
+
+    # If constant, return the y
+    return round(last_y / ENGINE_UNITS_PER_METER,0)
```
hassayag, line 204: nitpick, but this can be turned into an `int`. Otherwise every value has `.0` at the end

src/parser/parsers/heroes.py
```diff
@@ -202,3 +164,41 @@ def _map_attr_names(self, data, func):
             output_data[mapped_key] = data[key]
 
         return output_data
+    
+def calc_bullet_velocity(spline):
```
hassayag, line 168: I meant just change the name to `_calc_bullet_velocity`, not move it out of the class. It doesn't make much difference, but its good to keep it consistent with the rest of the functions

src/parser/parsers/heroes.py
```diff
@@ -107,57 +107,19 @@ def _parse_hero_weapon(self, hero_value):
...
-            'FalloffEndRange': weapon_prim['m_flDamageFalloffEndRange'] / ENGINE_UNITS_PER_METER,
+            'BulletSpeed': calc_bullet_velocity(weapon_prim['m_BulletSpeedCurve']['m_spline']),
+            'FalloffStartRange': round(weapon_prim['m_flDamageFalloffStartRange'] / ENGINE_UNITS_PER_METER,0),
+            'FalloffEndRange': round(weapon_prim['m_flDamageFalloffEndRange'] / ENGINE_UNITS_PER_METER,0),
             'FalloffStartScale': weapon_prim['m_flDamageFalloffStartScale'],
             'FalloffEndScale': weapon_prim['m_flDamageFalloffEndScale'],
             'FalloffBias': weapon_prim['m_flDamageFalloffBias'],
             'BulletGravityScale': weapon_prim['m_flBulletGravityScale'],
-            'BulletRadius': weapon_prim['m_flBulletRadius'] / ENGINE_UNITS_PER_METER,
+            'BulletRadius': round(weapon_prim['m_flBulletRadius'] / ENGINE_UNITS_PER_METER,0),
```
hassayag, line 117: looks like this is flattening every value to 0. Maybe 2/3 decimal places here? Also is this radius visible anywhere in game to make comparisons

hassayag (approved):
🚀

PR #19, opened 2024-09-16
Feature/attr map

Surxe:
Added localization route to attribute data; used by {{StatBoxes}}


src/parser/parsers/attributes.py
```diff
@@ -16,16 +16,77 @@ def __init__(self, heroes_data, localizations):
...
+
+        # Write the attributes to a json file
+        json_utils.write(OUTPUT_DIR + 'json/attribute-data.json', all_attributes)
+
+    def _map_to_unlocalized(self, all_attributes):
+        """Maps the attributes to their unlocalized names"""
+        """
+        Maps attributes as they appear in shop UI data, such as "BulletDamage" to their unlocalized names "StatDesc_BulletDamage"
+        The unlocalized name will then be localized on the front end
+        """
```
hassayag, line 51: Can these two comment blocks be combined

src/parser/parsers/attributes.py
hassayag, line 1: Since this file has more context, it would help to have a comment on the class describing how the output data is used. Maybe with a link to the statsbox on the wiki

hassayag (changes requested):
attribute-data has a nice layout. Just a couple small things

src/parser/parsers/attributes.py
Surxe: Good idea

src/parser/parsers/attributes.py
Surxe: done

src/parser/parsers/attributes.py
```diff
@@ -7,7 +7,10 @@
 import utils.json_utils as json_utils
 import maps as maps
 
-
+"""
+Output-data is used by https://deadlocked.wiki/Template:StatBoxes
+to display a hero's attributes on their hero page
+"""
```
hassayag, line 13: Sorry to be annoying, this should go below the class to be included in the docstring. I know I've been doing it wrong in some places, but I'm finding myfeet 😅

hassayag (approved):
🔥

src/parser/parsers/attributes.py
Surxe: ohhh I see, that makes sense actually. Done!

PR #24, opened 2024-09-19
Feature/level scaling; Melee split into light/heavy; MaxHealth>Health

Surxe:
Minor remappings so that it matches the attribute names that they scale


hassayag:
Needs a master merge before I can properly review. Also minor thing, let's use `bugfix` for now. `hotfix` is more associated with resolving immediate production issues.

Surxe:
> Needs a master merge before I can properly review. Also minor thing, let's use `bugfix` for now. `hotfix` is more associated with resolving immediate production issues.

complete

src/parser/parsers/heroes.py
```diff
@@ -184,4 +197,4 @@ def _calc_bullet_velocity(self, spline):
                 raise Exception('Bullet speed curve is not constant')
 
         # If constant, return the y
-        return last_y / ENGINE_UNITS_PER_METER
+        return int(round(last_y / ENGINE_UNITS_PER_METER, 0))
```
hassayag, line 200: Why is only this one rounded?

src/parser/parsers/heroes.py
```diff
@@ -81,23 +93,24 @@ def _parse_hero_weapon(self, hero_value):
...
-            'FalloffStartScale': weapon_prim['m_flDamageFalloffStartScale'],
-            'FalloffEndScale': weapon_prim['m_flDamageFalloffEndScale'],
-            'FalloffBias': weapon_prim['m_flDamageFalloffBias'],
-            'BulletGravityScale': weapon_prim['m_flBulletGravityScale'],
-            #'BulletRadius': weapon_prim['m_flBulletRadius'] / ENGINE_UNITS_PER_METER,
+            'BulletDamage': w['m_flBulletDamage'],
+            'RoundsPerSecond': 1 / w['m_flCycleTime'],
+            'ClipSize': w['m_iClipSize'],
+            'ReloadTime': w['m_reloadDuration'],
+            'ReloadMovespeed': int(w['m_flReloadMoveSpeed']) / 10000,
```
hassayag, line 103: oh wait, this also rounded

src/parser/parsers/heroes.py
Surxe: mistake with conflict merge; just scrapped this

src/parser/parsers/heroes.py
Surxe: in the data its string "10000", I want to confirm this later but I believe this just represents 1x multiplier

hassayag: Ah yeah it looks like it does nothing. In the case it does change, say10000 -> 14000, this would stil read out as 1, so we'd never know. So if its inconsequential could we just leave it as a float?

Surxe: oh derp, your right. just changed it from int(x)/10000 to float(x)/10000

hassayag (approved):

src/parser/parsers/heroes.py
hassayag: wait nah you're not a derp. I realise now the int is wrapping the value BEFORE it gets divided. So not really as bad as i thought. At the very least if they change it to `10000.5` it will "matter" 😂

PR #26, opened 2024-09-19
Lore, Playstyle, and Role keys from localization

Surxe:
Add localization keys to a hero's lore, playstyle, and lore texts


hassayag-equans (changes requested):
There is a large diff here. You probably need to rerun the parser

Surxe:
> There is a large diff here. You probably need to rerun the parser

Was missing changes from last merged PR, merged them and re-parsed output.

hassayag (approved):

PR #27, opened 2024-09-20
Feature/Ability UI parser to store format the parsed hero data into a usable format for display

hassayag:
- **Created `AbilityUiParser` to create a more usable format for the wiki ability cards**
- **Moved all localization-related logic from `AbilitiesParser` to `AbilityUiParser`**
- Moved most json write calls to the main parser


Surxe:
Is there a particular reason we need english localized output to the ability-ui.json? I would think that just the key to localization is sufficient.

For example, instead of

"Key": "ability_incendiary_projectile",
"Name": "Catalyst",
"Description": "Spew napalm that slows enemy movement and amplifies the damage Infernus does to them.",

I would recommend

"Key": "ability_incendiary_projectile", (up to you if this is still needed)
"Name": "ability_incendiary_projectile",
"Description": "ability_incendiary_projectile_desc"

This would have it follow the same format as hero-data.json's hero_key's WeaponName, WeaponDescription, Lore, Role, Playstyle, keys in WeaponTypes, etc.

Alternatively we can alter this slightly so it follows hero-data's pattern of allowing searching for an ability-ui object using a function similar to Module:HeroData's get_json_item, i.e., retrieve ability ui card with name "Catalyst" instead of "Infernus", "1".

For this we would want:
"NameKey": "ability_incendiary_projectile", (key for localizing name)
"Name": "Catalyst", (searchable by get_json_item)
"DescriptionKey": "ability_incendiary_projectile_desc", (key for localizing description)

To expand on this idea; you would use "XKey" for all localizable content other than the Ability's actual name, as thats realistically the only object we might want to be able to use get_json_item on.

Before simply appending _desc to the NameKey to get DescriptionKey, you may want to ensure its fully correct similar to attributes.py's _map_to_unlocalized which locates a key in any of the ways they exist. This part may not be needed if the key is 100% consistently just appending _desc.

hassayag:
Added loop to run ui parser for every language, writing any localization updates to their relevant file. This means any complex string regex replacements are run once on the backend

Surxe (approved):
Pushed some very minor changes, look through specifically the 3rd one and see if anything needs to occur there. Other than that, looks awesome! Really well organized, nicely done!

src/parser/parsers/ability_ui.py
```diff
@@ -158,8 +161,12 @@ def _parse_main_block(self, info_section, parsed_ability):
...
             title_key = props.get('m_strPropertiesTitleLocString')
             if title_key is not None:
-                # localization keys are prefixed with a "#"
-                title_key = title_key.replace('#', '')
+                # localization keys are prefixed with a "#", remove it
+                if title_key.startswith('#'):
+                    title_key = title_key[len('#'):]
+                else:
+                    # There are some cases of this
+                    title_key = title_key.replace('#', '')
```
hassayag, line 169: @Surxe I don't see how this is needed. Are there any cases where the `#` is not at the beginning?

Surxe: Its not needed currently, just guards against potential valve changes.

hassayag: Right, that's all good. But this implies that a future key would have a # in the middle of the word. I will just revert it to keep the code clearer

Surxe: Oh sorry I was referring to the first one I added where theres currently no exceptions. Your right for this one its a pointless addition

PR #28, opened 2024-09-20
Feature/attr alt name removed

Surxe:
Adds 2 hero stat remaps to remove the need for attributes to have alternate_name in the data


src/parser/parsers/heroes.py
```diff
@@ -197,4 +197,4 @@ def _calc_bullet_velocity(self, spline):
                 raise Exception('Bullet speed curve is not constant')
 
         # If constant, return the y
-        return last_y / ENGINE_UNITS_PER_METER
+        return int(round(last_y / ENGINE_UNITS_PER_METER, 0))
```
hassayag, line 200: hes back

Surxe: lol how does it do this, it never prompted me in merge conflict

hassayag (approved):

src/parser/parsers/heroes.py
hassayag: It should overwrite when you merge in master you'd think 🤔

PR #29, opened 2024-09-21
Feature/wep-attr-changes

Surxe:
- Weapon name, description, and types now parsed in hero parser
- Attributes data streamlined as the nearby weapon data is parsed via hero parser instead


src/parser/parsers/attributes.py
```diff
@@ -138,42 +115,29 @@ def _parse_shop_stat_display(self, hero_value):
...
+            # Add stat if not already present
+            if stat_mapped not in category_attributes:
+                category_attributes[stat_mapped] = {}
 
         return category_attributes
+
+
 """
 Output-data is used by https://deadlocked.wiki/Template:StatBoxes
 to display a hero's attributes on their hero page
```
hassayag, line 143: Has this always just been chillin at the bottom of the page?

hassayag (approved):

src/parser/parsers/attributes.py
Surxe: yeah, for a few PR's at least

hassayag: If its still relevant, can it not be on the class documentation?

PR #30, opened 2024-09-21
Feature/statbox attr list

Surxe:
List of attributes that are in the in game shop UI and are therefore used by wiki statboxes now outputted to Data:StatBoxAttrs.json


hassayag (approved):
🔥

PR #36, opened 2024-09-22
Feature/bullet per shot fix

Surxe:
Abrams shoots 9 bullet per left click, 9x'ing his dps


hassayag (commented):
As we add more and more weapon data, should we consider nesting it under "Weapon"? 

And thinking about the future, I have a feeling they will let you change weapons on a char, so could have a `DefaultWeapon` field that references `weapons.json`

Surxe:
> As we add more and more weapon data, should we consider nesting it under "Weapon"?
> 
> And thinking about the future, I have a feeling they will let you change weapons on a char, so could have a `DefaultWeapon` field that references `weapons.json`

Will look into this for future PR, as it will require a decently sized rework

hassayag (approved):

PR #40, opened 2024-09-24
Feature/hero comparison attr list

Surxe:
Form a list of non-constant stats that will be displayed by deadlocked.wiki/Hero Comparison, including SustainedDPS newly added stat


src/parser/parsers/heroes.py
```diff
@@ -84,7 +84,52 @@ def run(self):
...
+
+    def _write_non_constants_stats(self, all_hero_stats):
+        """
+        Writes list of non-constant stats to json file.
+
+        Returns non_constants_stats dict, where elements are true if they are non-constant.
+
+        Non-constant stats are ones that will be displayed
+        on the deadlocked.wiki/Hero_Comparison page, among others in the future.
+        """
```
hassayag, line 100: I think be clearer as to what non-constant means. Eg. `a "non-constant" stat is one that differs between heroes and is therefore suitable to be displayed on the Hero_Comparison page`

hassayag (approved):
Looks good. I do worry that it requires too much human intervention to get the data out there. Perhaps a future PR could move that array from the lua script to somewhere in this repo. And we can have CI block the merge if the table is missing an element. Then it can categorically never be missed

src/parser/parsers/heroes.py
```diff
@@ -84,7 +84,52 @@ def run(self):
 
                 all_hero_stats[hero_key] = json_utils.sort_dict(hero_stats)
 
-        return all_hero_stats
+        # Write non-constant stats to json file
+        non_constants = self._write_non_constants_stats(all_hero_stats)
+
+        return all_hero_stats, non_constants
+
+    def _write_non_constants_stats(self, all_hero_stats):
```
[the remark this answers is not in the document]
Surxe: Fixed! Previous implementation did write a json, had forgotten to rename it.

src/parser/parsers/heroes.py
Surxe: Renamed from "non-constant-stats" to "meaningful-stats"

Surxe:
Noticed that it wasn't factoring in stats that are initially constants, but have scaling on some heroes. Going to add these then re-request review.

Surxe:
> Noticed that it wasn't factoring in stats that are initially constants, but have scaling on some heroes. Going to add these then re-request review.

Complete!

src/parser/parsers/heroes.py
hassayag: Still requires an explanation as to what meaningful or non-constant means

output-data/json/hero-non-constants.json
hassayag, line 1: delete

hassayag (changes requested):
Needs reparsing, looks to be out of date

src/parser/parsers/heroes.py
Surxe: added

Surxe:
Re-merged develop into branch, decompiled most recent patch, and re-parsed. Should be all good to go now.

src/output-data/json/ability-data.json
hassayag, line 1: i think this is a remnant from the bad env var. delete

Surxe: > i think this is a remnant from the bad env var. delete

not sure what you mean, ability-data.json is referenced by parser.py and changelogs.py

hassayag: The directory is not correct

output-datajson/hero-meaningful-stats.json
hassayag, line 1: another dodgy file path

Surxe: nice catch hah, fixed!

src/output-data/json/ability-data.json
Surxe: Ooooh I see what you mean now, okay fixed this as well

Surxe:
Remerged develop branch

output-data/version.txt
```diff
@@ -1,9 +1,9 @@
...
+ServerVersion=5220
 ProductName=citadel
 appID=1422450
 ServerAppID=1422460
 ToolsAppID=211
-SourceRevision=9219272
-VersionDate=Sep 30 2024
-VersionTime=11:12:14
+SourceRevision=9217762
+VersionDate=Sep 27 2024
```
hassayag, line 8: I put the new patch on develop so should reparse and then its good to go!

Surxe: I already reparsed, looks like it matches develop

hassayag: this is a diff between your branch and develop. Maybe steam needs to update the game?

hassayag (approved):
🚀

PR #42, opened 2024-09-25
Bugfix/pandas dataframe in develop branch

B13rg:
Getting warnings about `df.map` vs. `df.applymap`.
Right now it's just a warning, once we update pandas to `2.1.0`, it will need to be updated.

ref: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.applymap.html


hassayag (approved):
Nice, good spot 👌

PR #48, opened 2024-09-28
Feature/ability description formatting

hassayag:


Surxe (approved):
presume the only out-out-date bits from develop branch is the last merged PR.

Looks goods!

*edit, looks I just needed to refresh, now it says its up to date

PR #51, opened 2024-10-02
Feature/decompile py

B13rg:
Moves decompile process from a shell script to part of deadbot.

* added cmdline parameters that also read env vars for setting variables
* remove cli interface from python files in decompile folder
* add `-d` `--decompile` flag and `-b` `--bot_push` flags


src/deadbot.py
```diff
@@ -37,6 +40,71 @@ def _update_page(self, page):
...
+        "-d",
+        "--decompile",
+        action="store_true",
+        help="Decompiles Deadlock game files. (also set with DECOMPILE environment variable)",
+    )
+    parser.add_argument(
+        "-b",
+        "--bot_push",
+        action="store_true",
+        help="Push current data to wiki (also set with BOT_PUSH environment variable)",
```
hassayag, line 88: can this default to `BOT_PUSH` env var?

src/deadbot.py
```diff
@@ -1,6 +1,9 @@
 import os
 import mwclient
+import argparse
+# local imports
```
hassayag, line 4: delete comment

src/deadbot.py
```diff
@@ -37,6 +40,71 @@ def _update_page(self, page):
...
+    args = parse_arguments()
+
+    if args.decompile or os.getenv("DECOMPILE", False):
+        print("Decompiling source files...")
+        decompile.decompile(
+            args.dl_path, args.workdir, args.output, args.decompiler_cmd
+        )
+    else:
+        print("! Skipping Decompiler !")
+
```
hassayag, line 104: I presume you meant to have the parser called here? That would also mean a corresponding operational flag

src/deadbot.py
```diff
@@ -37,6 +40,71 @@ def _update_page(self, page):
...
+        help="Push current data to wiki (also set with BOT_PUSH environment variable)",
+    )
+
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = parse_arguments()
+
+    if args.decompile or os.getenv("DECOMPILE", False):
```
hassayag, line 97: take care with env vars as booleans. I'm seeing my `"false"` env var being evaluated as truthy and therefore running these steps regardless

src/deadbot.py
```diff
@@ -37,6 +40,71 @@ def _update_page(self, page):
...
+        help="Command for Valve Resource Format tool (also set with DECOMPILER_CMD env variable)",
+        default=os.getenv("DECOMPILER_CMD", "tools/Decompiler"),
+    )
+
+    # Operational Flags
+    parser.add_argument(
+        "-d",
+        "--decompile",
+        action="store_true",
+        help="Decompiles Deadlock game files. (also set with DECOMPILE environment variable)",
```
hassayag, line 82: Shall we default to use `DECOMPILE` here, similar to the other flags. As I noted in another comment, make sure you do checks for string bools

hassayag (changes requested):
I appreciate you doing the not-so-sexy work of tidying up the repo 🫡

Few comments to deal with, but its so much cleaner!

src/deadbot.py
B13rg: rm'd

src/deadbot.py
B13rg: Eventually, I have some changes that "parameterize" the rest of the scripts like that and the s3 stuff, 

It'll be in a pr right after this one so it isn't overwhelming

src/deadbot.py
B13rg: Woops yep!  It just checked if it existed, not what it was.
Added a match against `[True, 'true', 'True', 'TRUE', 't', 'T', 1]`

B13rg: For the "action" parameters like decompile, I have them setup as flags so you can do `-d` instead of `-d=1`.
Because of that, the environment variables for those params are checked at runtime, which you found the bug in below!

It's setup so the user can use the parameter or env var, or a default value if one is sensible.
A truth table looks like this:
||env t|env f|
|param t| t| t|
|param f|t | default |

For all the "action" type flag params, they default to false or none.
For set parameters that have values, they default to hopefully good defaults

src/deadbot.py
B13rg: Woops yep!  It just checked if it existed, not what it was.
Added a match against `[True, 'true', 'True', 'TRUE', 't', 'T', 1]`

src/deadbot.py
B13rg: Yep!  It will default to that var when it checks for the parameter and it didn't "store_true".
If the env var isn't set, it'll default to false

src/deadbot.py
```diff
@@ -37,6 +40,71 @@ def _update_page(self, page):
...
+        decompile.decompile(
+            args.dl_path, args.workdir, args.output, args.decompiler_cmd
+        )
+    else:
+        print("! Skipping Decompiler !")
+
+    if args.bot_push or os.getenv("BOT_PUSH", False):
+        print("Running DeadBot...")
+        bot = DeadBot()
+        bot.run()
```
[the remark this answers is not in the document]
B13rg: aaaand changed.  Renamed `bot.run()` -> `bot.push_lane()`

main.sh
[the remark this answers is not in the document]
B13rg: Nope!  My goal is to move all the main.sh logic into deadbot, and then edit main.sh to have examples of how to run it.
My first try I outright removed it, but I think it's still good for reference on how to do different things with `Deadbot`

src/deadbot.py
hassayag-equans: Ah i get you, this is just for decompilation. Hence keeping main.sh until its fully migrated

hassayag: Oh I get you, this is just for decompilation, and once the parser is added we can deprecate main.sh 👌

hassayag (approved):
🚀

hassayag:
Sorry just realised this was targetting master, but shouldn't be an issue as I just released a new version to master

PR #54, opened 2024-10-03
Feature/Changelog RSS Fetch

B13rg:
Currently, getting changelogs requires manually updating a changelog file and running `Deadbot`.
This PR adds a way to fetch the changelogs from the forum rss feed.

The changelog notes are posted at https://forums.playdeadlock.com/forums/changelog.10/ , and the forum rss feed is https://forums.playdeadlock.com/forums/changelog.10/index.rss

I use `feedparser` and `std.HTMLParser` to fetch the feed, extract the changelog content, and send it through the same process as the text files.

This PR includes changes from the latest release (10-02-2024 update)


src/parser/parsers/changelogs.py
```diff
@@ -11,27 +13,60 @@
...
-            changelog = self.run(date)
-            changelogs_by_date[date] = changelog
+            changelog = self.run(date, self._read_local_logs(date))
+            self.changelogs_by_date[date] = changelog
+
+    def run_all(self, fetch_rss=False):
+        self.changelogs_by_date = {}
+        if fetch_rss:
+            self.fetch_forum_changelogs()
+        self.process_local_changelogs()
```
hassayag, line 63: So to test the fetched logs, I deleted all raw-changelogs and ran the parser. It seems there is some limit to the fetching, since there are massive chunks of info missing

src/parser/parsers/changelogs.py
```diff
@@ -11,27 +13,60 @@
...
+        f_url = 'https://forums.playdeadlock.com/forums/changelog.10/index.rss'
+        # fetches first page, last 20 entries date-wise
+        feed = feedparser.parse(f_url)
+        for entry in feed.entries:
+            # Dependent on thread title being in the format "mm-dd-yyyy Update"
+            date = entry.title.replace(' Update', '')
+            html_remover = HTMLTagRemover()
+            # Potential for more than one post, we just grab the first
+            html_remover.feed(entry.content[0].value)
+            self.changelogs_by_date[date] = self.run(date, html_remover.get_text())
```
hassayag, line 50: I would add a step to save the raw changelog along with the others, as its nice to keep our own archive

src/parser/parsers/changelogs.py
```diff
@@ -11,27 +13,60 @@
...
         self.CHANGELOGS_DIR = os.path.join(os.path.dirname(__file__), '../raw-changelogs/')
         self.OUTPUT_DIR = OUTPUT_DIR
-        self.OUTPUT_CHANGELOGS = self.OUTPUT_DIR+'/changelogs'
+        self.OUTPUT_CHANGELOGS = self.OUTPUT_DIR + '/changelogs'
         self.resources = self._get_resources()
 
-    def run_all(self):
-        changelogs_by_date = {}
+    # download rss feed from changelog forum and parse entries
+    def fetch_forum_changelogs(self):
```
hassayag, line 38: this whole fetching process feels like it could be its own file and class. Thoughts on a `parsers/changelog` folder which contains `changelog_parser.py` and `fetch_changelogs.py`?

src/parser/parser.py
```diff
@@ -88,7 +88,8 @@ def _merge_localizations(self, language, group, data):
             # duplicate key error. This is a temporary measure to keep patch updates going
             elif group != 'heroes':
                 current_value = self.localizations[language][key]
-                raise Exception(
```
hassayag, line 91: Should this be changed? Genuinely not sure what to do about it

hassayag (changes requested):
Good stuff! I have concerns about the consistency of the imported changelogs, as some are being cut off. If that can be ironed out then this is one fewer thing to do manually each patch

src/parser/parsers/changelogs.py
B13rg: Ahh, I think it limits it to 20 per page, I'll add a pagination loop

src/parser/parser.py
B13rg: Yeah, when I have it replace it print it out correctly, I gave a clean try at is here: #66

src/parser/parsers/changelogs.py
B13rg: So the rss feed will only show the latest 20, no paging.  It won't capture the old ones, but it'll at least catch everything moving forward :/

src/parser/parsers/changelogs.py
B13rg: Done!  I created a "raw-changelogs" folder in the output directory, and have rss and txt subfolders that contain the changelogs

src/parser/parser.py
B13rg: Now I'm not sure..... I'll revert it, it's probably a linux thing

src/parser/parsers/changelogs.py
B13rg: Separated them!  Now there's a `parse_changelogs` and a `fetch_changelogs`.  `fetch_*` gathers the txt and rss stuff, and prepares the logs.  It will load logs from rss, then the txt files.  With any duplicate changelog titles, the txt logs will be used instead of the fetched one.  This will make it so we can still customize the changelog text in case manual editing is required for some reason

B13rg:
Okay, should be good to go!

src/parser/parsers/changelogs/fetch_changelogs.py
```diff
@@ -0,0 +1,67 @@
...
+        for div in soup.find_all('div', class_='bbWrapper'):
+            entries.append(div.text.strip())
+        return entries
+
+    # download rss feed from changelog forum and parse entries
+    def fetch_forum_changelogs(self):
+        print('Parsing Changelog RSS feed')
+        # fetches 20 most recent entries
+        feed = feedparser.parse(self.RSS_URL)
+        out_dir = self.OUTPUT_CHANGELOGS + '/rss'
```
hassayag, line 35: I think we just need one directory for raw changelogs (current raw-changelogs dir), not three. When the parser runs, it will:
1. Query rss feed to get the latest changelog dates
2. See if any are missing in raw-changelogs
3. If they are missing, query for them and writes to the raw-changelogs dir

src/parser/parsers/changelogs/parse_changelogs.py
```diff
@@ -1,37 +1,27 @@
...
-        changelogs_by_date = {}
-        files = [f for f in listdir(self.CHANGELOGS_DIR) if isfile(join(self.CHANGELOGS_DIR, f))]
-        for file in files:
-            date = file.replace('.txt', '')
-            changelog = self.run(date)
-            changelogs_by_date[date] = changelog
-
+    def run_all(self, txt=True, rss=True):
+        log_fetch = ChangelogFetcher(self.OUTPUT_DIR) 
+        for key, val in log_fetch.fetch_changelogs().items():
```
hassayag, line 18: `key, val` is too generic, I think `date, changelog` fits

src/parser/parsers/changelogs/parse_changelogs.py
```diff
@@ -1,37 +1,27 @@
...
 
-    def run_all(self):
-        changelogs_by_date = {}
-        files = [f for f in listdir(self.CHANGELOGS_DIR) if isfile(join(self.CHANGELOGS_DIR, f))]
-        for file in files:
-            date = file.replace('.txt', '')
-            changelog = self.run(date)
-            changelogs_by_date[date] = changelog
-
+    def run_all(self, txt=True, rss=True):
```
hassayag, line 16: unused arguments

hassayag (changes requested):

src/parser/parsers/changelogs/parse_changelogs.py
B13rg: Fixed! Moved whether is parses txt or rss logs moved outside of that class.
Now it just goes through `self.changelogs_by_date`

src/parser/parsers/changelogs/parse_changelogs.py
B13rg: renamed!

src/parser/parsers/changelogs/fetch_changelogs.py
```diff
@@ -0,0 +1,67 @@
...
+            os.makedirs(out_dir, exist_ok=True)
+        for file in files:
+            date = file.replace('.txt', '')
+            with open(self.CHANGELOGS_DIR + f'{date}.txt', 'r', encoding='utf8') as f:
+                changelogs = f.read()
+                self.changelogs_by_date[date] = changelogs
+                with open(out_dir + f'/{date}.txt', 'w', encoding='utf8') as f_out:
+                    f_out.write(changelogs)
+
+    def fetch_changelogs(self):
```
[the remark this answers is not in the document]
B13rg: Fixed!

src/parser/parsers/changelogs/fetch_changelogs.py
B13rg: I ended up created an `input-data` directory, and placed the raw changelogs in there.  I noticed while running it there are some syntax and spelling fixes in the changelogs, which I tried to preserve.  It works similar to what you outlined:
1. queries rss, get's latest posts, and fetches each posts content and stores is keyed on date.  
2. goes through each file in raw-changelogs, and adds / replaces the changelog if it already existed. (good fixing changelog text for reasons)
3. takes combined dict of changelogs, write to output files
4. passes changeslogs to to the parser, which extracts out the final specific changelog files.

src/parser/parsers/changelogs/parse_changelogs.py
```diff
@@ -1,67 +1,22 @@
...
-    def get_text(self):
-        return ''.join(self.result)
-
 
 class ChangelogParser:
-    def __init__(self):
+    def __init__(self, output_dir):
         self.CHANGELOGS_DIR = os.path.join(os.path.dirname(__file__), '../raw-changelogs/')
-        self.OUTPUT_DIR = OUTPUT_DIR
+        self.OUTPUT_DIR = output_dir
```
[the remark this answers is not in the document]
hassayag: I like what you've done with this, simpler to keep env reads in a single place 👌

src/parser/parsers/changelogs/fetch_changelogs.py
hassayag: Ah good idea with the input data, didn't consider fixing changelogs at this point, but that's easier than once its parsed.

I'd say all you need is a check for whether it exists in output-data already, as querying every time has no use if its already been done once.

I can see you've added a `CHLOGS` flag to maybe mitigate this, but the same issue will come up when you are trying to develop on the changelog parser. So I'm not sure I see a need for this flag, happy to be proven wrong though

hassayag (changes requested):

src/changelogs/parse_changelogs.py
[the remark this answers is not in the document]
B13rg: I wanted it a bit more separate, where it can be ran independently of the game file parsing, to possibly get the release notes up if not the parsed gamefile changes

README.md
```diff
@@ -17,15 +17,96 @@ Some data has been processed and formatted for google sheets:\
     2. Add to ~/.bash_profile - `export PATH="$PATH:<python_path>/Python311/Scripts"`
 
 3. `python3 -m pip install poetry`
-4. `python3 -m poetry install`
+4. `poetry install`
```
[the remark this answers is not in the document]
B13rg: Fixed!

src/changelogs/parse_changelogs.py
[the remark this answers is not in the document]
hassayag: Ah, so would you say this is just until the full end-to-end process is implemented for parsing?

hassayag:
if this is still a WIP ignore me, but this is the last outstanding item

> I'd say all you need is a check for whether it exists in output-data already, as querying every time has no use if its already been done once.

src/parser/parsers/changelogs/fetch_changelogs.py
B13rg: Yeah, I expanded it out to be "CHANGELOGS", and got rid of the SKIP_RSS flag
Now, It'll fetch the rss like usual but if a changelogs entry already exists, it won't re-fetch them.  this is done by loading in changelogs from the output directory first, then the rss feed, then the txt input directory.

B13rg:
Okay, set it up so it'll feed the existing output data in first, and then skip any rss items that already exist.

src/changelogs/fetch_changelogs.py
```diff
@@ -0,0 +1,69 @@
...
+class ChangelogFetcher:
+    def __init__(self):
+        self.changelogs_by_date = {}
+
+    def get_rss(self, rss_url, update_Existing = False):
+        self.RSS_URL = rss_url
+        self._fetch_forum_changelogs(update_Existing)
+        return self.changelogs_by_date
+
+    def get_txt(self, changelog_path, update_Existing = True):
```
hassayag, line 18: unused argument

src/changelogs/fetch_changelogs.py
```diff
@@ -0,0 +1,69 @@
...
+import feedparser
+from bs4 import BeautifulSoup
+from urllib import request
+
+
+class ChangelogFetcher:
+    def __init__(self):
+        self.changelogs_by_date = {}
+
+    def get_rss(self, rss_url, update_Existing = False):
```
hassayag, line 13: Should `update_Existing` have a capital "E"?

src/utils/parameters.py
```diff
@@ -117,7 +111,7 @@ def load_arguments():
     if not args.parse:
         args.parse = os.getenv('PARSE', False)
     if not args.changelogs:
-        args.changelogs = os.getenv('CHLOGS', False)
+        args.changelogs = os.getenv('CHANGELOGS', False)
```
hassayag, line 114: You're no longer a hot shot dev by using the full word smh 😂

hassayag (approved):
Lookin good! Just a couple minor changes to sort out. You can merge when those are done 👌

src/changelogs/fetch_changelogs.py
B13rg: fixed!

src/changelogs/fetch_changelogs.py
B13rg: removed it, it can always be added and for now the files someone creates should always be loaded

B13rg:
Woohoo!!!! Thank you for helping get through it, that was a lot of changes and commits!

PR #55, opened 2024-10-03
Feature/oop resource storage

Surxe:
See later comments, this post is outdated. (sry, couldn't find strikethrough, and ~~ doesn't seem to work in this markdown)

Resources are instead stored as class objects when in memory, in a static hash called "objects". When the class is imported, saveObjects() can be used to store data in its current format into the "objects" hash. When the class is imported in another module that needs it, loadObjects() can be used to load them to memory once, removing the need for handling parameter passing and receiving.

Also note that ability-ui is not moved over yet

Current draft only includes resource class and the classes that extend it, along with a brief example in item_parser.py, here
https://github.com/deadlock-wiki/deadbot/blob/45ec53c8962da96c6768fc196ea26a1f859913de/src/parser/resources/item/item_parser.py#L110


src/parser/resources/item/item_parser.py
```diff
@@ -101,12 +102,18 @@ def run(self):
...
             all_items[key] = parsed_item_data
 
         chart = MermaidDiagram(title='Items', nodes=self.nodes, links=self.links)
 
+        # Create and save Item objects
+        Item.hashToObjs(all_items)
+        Item.saveObjects(free_memory=True)
+
+        #Item.loadObjects() #if commented out, the following line will throw an error
+        print(Item.objects['armor_upgrade_t4'].getProp('Cost'))
```
B13rg, line 115: Should this be commented out as well?

Surxe: > Should this be commented out as well?

I left it uncommented in the draft before I got back to it today so you could see the error, but yes in order to run you would either make #115 a comment or uncomment #114

Surxe:
Finished and ready for review.

To summarize, there are **two main purposes** to this PR:
- Moving up an abstraction layer to make resource data parsing and storage easier to maintain moving forward
- Adds the capability to import the respective resource's class for smoother read/writing to/from memory

**Usage**
When needing to work with i.e. `parsed_abilities`, instead of passing it as parameters in all the intermediary steps, use `Ability.save_objects()` to save from memory to file, then where its needed, import Ability class and run `Ability.load_objects()`

All the added classes extend the Resource class, making it highly adaptable.
- Localization class overloads save and load methods as its saved to file per-language rather than just 1 file
- Item class adds a method for saving the mermaid diagram chart for component tree
- Attribute class adds a method for saving `meaningful_stats` and `order_lists`
- All the other resources don't yet have any warranted features that the Resource class doesn't have, though that will definitely change

Changelog's will be undergoing a restructure fairly soon, so don't need to put too much stock into that area.

PR #50 will be removed as its rendered unnecessary with this.

Appreciate any feedback!

hassayag (changes requested):
This is a cool! I need more time to digest it and will be quite busy this weekend, so I'll leave some initial thoughts for now.

- As you said, this intends to reduce the number of argument passing, which I like in theory, but i have some issues:
  - It assumes the data has been parsed already. Eg. if heroes is parsed after abilities, abilities cannot use that resource. This is where arguments make that relationship clear
  - There only seems to be a couple arguments fewer. Should this extend to localizations, raw data, etc?

- I don't quite like the use of `hash` so liberally. Eg. `hash_to_obj` should probably just be named `store_resources`, since it doesn't take in a hash, and doesn't return anything
- There is a massive diff on output-data

I like this direction, I can see this leading to having proper schemas for each resource, as right now it is totally untyped

Surxe:
>   * It assumes the data has been parsed already. Eg. if heroes is parsed after abilities, abilities cannot use that resource. This is where arguments make that relationship clear
Correct! We wouldn't want to make a cycle, i.e. parsing abilities requires heroes *and* parsing heroes requires abilities. This shouldn't ever be necessary, as passing (raw abilities + raw heroes) to both the hero_parser and abilities_parser can accomplish this. I discuss it a bit more in the next bullet point.

>   * There only seems to be a couple arguments fewer. Should this extend to localizations, raw data, etc?
**raw data**
We *could* extend it to raw data, though I think it better just for parsed data, at least in its current design. Would keep the code a lot cleaner for resources this way. Another option is to create another set of classes that extend "RawResource", but seems a bit over-abstract at that point. If you felt strongly about this, you could definitely convince me to implement it, but it doesn't need to be in this PR.

**localizations**
It does extend to localizations, though it is stored a bit differently; the Localization class overrides both load and save functions from Resource class since each localization object has its own file, unlike other resources. See `output-data\resources\localizations`.

> * I don't quite like the use of `hash` so liberally. Eg. `hash_to_obj` should probably just be named `store_resources`, since it doesn't take in a hash, and doesn't return anything
Done! Will leave `objs_to_hash` as is for now as mentioned in the comment its just a placeholder until Changelog restructure.

> * There is a massive diff on output-data
**Changelogs**:
After an unfortunate duration of 30 minutes of looking into it, it seems to be because the resource's are ordered differently in the output now. There was no `sort_dict` applied to them before. The current structure of changelog parser works in a way such that each line can only have 1 "tag" (as we referred to them more recently), which impacts whats removed from the prefixed semicolon in the line, and the {{AbilityIcon}} call thats written. Here's an example:

Spiritual Overflow / upgrade_tech_overflow: line 5880 to line 3018
Spirit Lifesteal / upgrade_health_stealing_magic: line 3001 to line 3182

Before, Spirit Lifesteal was first, now, Spiritual Overflow is first. As a result, their output changes
image

See the file diff here: https://github.com/deadlock-wiki/deadbot/pull/55/files#diff-8efdc146f801b6f836cc78372a01ffbbd6d64820105670e3c88a4fb7f6cd27e0L5

I'm not exactly certain on what's shuffling the resource objects, though I've applied `sort_dict` to the resources in the changelog parser now in commit `sort_dict for resources in changelog parser`, where you can see how it affects the changelog files in a similar fashion to as outlined above.

What is your opinion on deleting changelog files for now from `/develop` branch via PR? Since they aren't being used right now and the drafted changelog restructure is next on my platter, it seems like no downside for cleaning up the "files changed" section on this PR from 100 to like, <20 ish.

**Localization**
Just confirmed these are out of date on /develop branch. As I was double checking this I saw there was a steam update today, it changed my version.txt but nothing else changed after re-decompiling on this branch. Not sure what's out of date on /develop branch but it generates those localization diffs just the same when I decompile over there currently.

> I like this direction, I can see this leading to having proper schemas for each resource, as right now it is totally untyped
For sure! The original plan was to have these class schemas fleshed out, but I saw the absurd number of keys required for ability data and had to scrap it for at least this PR. Though, since you pointed it out and I thought on it more today, I'm going to *try* implementing a method in which I can specify important attributes for each object, and if any aren't specified, it will just go to the current `.data[key]` attribute. In the future, we should look into categorizing these attributes a bit better so we can have them in the schema more clearly. For example 'weapon stats' from hero data can be clustered under a 'Weapon' attribute. Will work on that this weekend and comment back when that's ready to be re-reviewed.

Thanks for the feedback!

Surxe:
Also forgot to mention but I was unable to decompile until recently, Brendan's bugfix corrected it so I merged that in and re-decompiled. Sadly didn't lower that "files changed" number lol, but thought worth mentioning regardless.

Surxe:
> For sure! The original plan was to have these class schemas fleshed out, but I saw the absurd number of keys required for ability data and had to scrap it for at least this PR. Though, since you pointed it out and I thought on it more today, I'm going to _try_ implementing a method in which I can specify important attributes for each object, and if any aren't specified, it will just go to the current `.data[key]` attribute. In the future, we should look into categorizing these attributes a bit better so we can have them in the schema more clearly. For example 'weapon stats' from hero data can be clustered under a 'Weapon' attribute. Will work on that this weekend and comment back when that's ready to be re-reviewed.

Okay, think I've got everything for this complete. Refactored save, load, and store such that they now more readily support customizing the attributes in a class definition. If any keys are added to it via store_resources that are not an attribute, it goes under the 'Data' attribute. 

I added attributes to all the objects wherever currently plausible, though before its fully ready we need to discuss how to go about grouping data under more reasonable attributes for each class:

**AbilityUI**

In AbilityUI.json, the only keys are Key, Name (hero), 1, 2, 3, and 4. Ofc, we can't use numbers as the name of the attribute. What are your thoughts on the 1st layer of this output being the ability key instead of the hero key? BoundAbilities in the hero data would then be changed such that each ability has just the key and not the actual data in it, i.e.
BoundAbilities = {
  1: ability_incendiary_projectile
  2: ability_flame_dash
  3: ability_afterburn
  4: ability_fire_bomb
}

AbilityUI.json would then look closer to the current Ability.json output, such that the first layer is the ability key which refers to this BoundAbilities area instead, that way theres no repeated information. We would use Hero.BoundAbilities to determine the # ability that the hero uses it in.

**Attribute**
I may change the Attribute.json structure down the line, though its not necessary to right now unlike the AbilityUI.

**Hero**
I would also work on grouping Weapon related keys under a 'Weapon' attribute in hero data, the rest should be under the 'Data' attribute.

**Changelog**
Changelog can be ignored for now.

**Item and Ability**
Item and Ability are complete, though they will require a few tweaks on frontend.

**To summarize, I'd like to group Weapon stats in hero data, and remove first 2 layers of AbilityUI data, before merging. What are your thoughts?**

B13rg (approved):
Looks good, I got clean runs from it
I like the added sorting as well!

src/parser/resources/item/item_objects.py
```diff
@@ -0,0 +1,32 @@
...
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '../..')))
+import resources.resource as resource
+
+
+# extends Resource
+class Item(resource.Resource):
+    objects = {}
+    chart = None
+
+    def __init__(self, key):
```
hassayag, line 13: To me it makes more sense if each resource takes in the data immediately, instead of happening outside of the class. That keeps it self contained, and allows us to have any specific validation if we want

hassayag (changes requested):
I have some concerns for the architecture. The way I would improve this is to have an object, repository, and parser.  

Resource Objects - initialises that resource
Repository - a singleton class instance that stores all resource objects for that type. Also would handle all saving and loading
Parser - uses the repositories to trigger read/writes to the repository and ultiamtely to json files

I would recommend splitting this PR into smaller PRs, one for each resource type, as it makes iterating on it much quicker, plus its easier to review. So let's start with items and get that polished before moving onto the others

src/parser/resources/hero/hero_parser.py
```diff
@@ -2,17 +2,21 @@
 import os
 
 # bring utils module in scope
-sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '../..')))
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..')))
```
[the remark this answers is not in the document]
Surxe: I'm not sure how, but a repository hosted by a friend for previous wiki found a way for all directories to be added to sys path without any extra work handling it. I'll ask around.

src/parser/resources/item/item_objects.py
```diff
@@ -0,0 +1,32 @@
...
+        self.Description = ''
+        self.Cost = ''
+        self.Tier = ''
+        self.Activation = ''
+        self.Slot = ''
+        self.Components = []
+        self.TargetTypes = []
+        self.ShopFilters = []
+        self.Disabled = True
+        self.Data = {}
```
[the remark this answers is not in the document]
Surxe: its temporary storage for all keys that are yet to be grouped correctly.

Surxe:
Closing this PR. Too much work for me to take on along with other more relevant projects for the wiki. Maybe it can be revisited in the future. Note to self for future wiki projects, don't alter the structure of data so much, the way it was grouped in decompiled data was honestly very impressive of Valve, the way we have it grouped in parsed data currently feels unnecessary, and undoing this is what makes the PR take too much time.

PR #66, opened 2024-10-04
Bugfix/Fix develop branch deadbot call

B13rg:
Makes the following changes:

* fix how `main.sh` calls various deadbot functions
* added parenthesis around parameter logic
* soften localization exception

### DIFF
--- a/main.sh
+++ b/main.sh
@@ -8,7 +8,7 @@ fi
 
 if [ "$DECOMPILE" = true ]; then
     cd src
-    python3 deadbot.py --decompile=true --deadbot=false
+    python3 deadbot.py --decompile
     cd ..
 fi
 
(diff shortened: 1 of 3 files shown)


B13rg:
For the `parser.py` changes, I kept hitting it (maybe a python 3.12 thing?) so I softened it to a warning.  With the assignment added, I get a clean diff when ran against files

hassayag:
> For the `parser.py` changes, I kept hitting it (maybe a python 3.12 thing?) so I softened it to a warning. With the assignment added, I get a clean diff when ran against files

Ah I am on 3.11 and so is the project in general, see if that helps? But not sure why that would be a problem in this case 🤔

B13rg:
Actually, I think the changes in https://github.com/deadlock-wiki/deadbot/pull/53 will fix this better for `main.sh`, it finishes parameterizing all the functions and cleaning up that interface.

PR #69, opened 2024-10-10
Small bugfixes to output folders and docker image

B13rg:
* fix docker-compose image 
* Add safety around opening directories
* Stopped json compare from throwing exception, just returns false


src/changelogs/fetch_changelogs.py
```diff
@@ -57,6 +57,10 @@ def _fetch_forum_changelogs(self, update_existing = False):
 
     def _process_local_changelogs(self, changelog_path):
         print('Parsing Changelog txt files')
+        # Make sure path exists
+        if not os.path.isdir(changelog_path):
+            print(f'Issue opening changelog dir `{changelog_path}`')
+            return
```
hassayag, line 63: Is there a reason this shouldn't raise an error?

B13rg: Since the changelog directory is harcoded, there's always the change it doesn't exist.
Instead of crashing out, I just want it to alert the user and continue.

This is also for cases where there isn't any manual changes being passed and and the user just wants the changelogs.
It was an issue I was running into with automation in the deadlock-data repo

hassayag: Ah fair enough, thanks for calrifying!

hassayag (approved):

PR #70, opened 2024-10-10
Feature/UTF-8 encoding support for JSON read/write operations

cycleapple:
This pull request addresses and improves the handling of JSON file reading, writing, and processing to ensure that Chinese localization data is parsed correctly. The main changes include:

### Added UTF-8 Encoding Support:
All file read and write operations now explicitly use UTF-8 encoding. This change is essential for handling non-ASCII characters (e.g., Chinese characters) without causing encoding errors or generating Unicode escape sequences in the JSON files.
### Enhanced Unicode Handling in process_files:
Modified the process_files function to use json.dump with the ensure_ascii=False parameter, which prevents non-ASCII characters from being escaped. This ensures that Chinese characters and other non-ASCII text are correctly displayed in the generated JSON files.
### Improved File I/O Operations:
Added encoding='utf-8' to all open statements to prevent potential UnicodeDecodeError and ensure compatibility with various language text files.
Improved error handling and added comments to clarify code behavior, making it easier to understand and maintain.
## Reason for Changes:
These changes are necessary to ensure that Chinese localization data is correctly parsed and displayed, providing a better experience for users viewing localized content. The previous implementation caused Chinese characters to be escaped, making it difficult to read and understand the content in the JSON files.

## Impact of Changes:
Improved Readability: Chinese and other non-ASCII characters will be correctly displayed in JSON files.
Increased Compatibility: Files with non-ASCII characters can be read and written without causing encoding errors.


Surxe:
Glad to have someone familiar with encoding on the team! I'm not able to review/test as I currently only have access to a crappy laptop during hurricane Milton. Anyways, does this also fix the issue seen below, where end lines aren't seen as end lines?
image

cycleapple:
I'm not an expert on this topic either—I was just trying to fix the Chinese parsing issue, which is pretty common with Chinese characters.

The `\n` issue seems to be caused by how line breaks are getting parsed and displayed. It’s probably not related to the encoding, so it’d be better to open another issue ticket for it. Also, having a way to reproduce the problem would be helpful.

cycleapple:
i encounter the same \n problem aprear in my tchinese result as well. let me try to fix the parser

cycleapple:
the json parse issue should fix separately in another PR

hassayag (changes requested):
@cycleapple Great to have you on here! Looks like this works perfectly, and has resolved issues not just related to Chinese parsing.

The only thing left is to run `main.sh` on the latest game files and pushing the updated localization files. Let me know if you run into any issues with that

Surxe:
@cycleapple let me know if you'd like me to run `main.sh` for you

Surxe:
Ran main.sh

hassayag (approved):
Looks good, just reparse on latest version and it should be good (words I hope to stop saying with the new pipeline 🤞)

PR #78, opened 2024-10-15
Bugfix/fix manual override of localized ability data

hassayag:
- **Added retrieval of localized token overrides from the game files**
- **Harsher error handling of "UNKNOWN" to prevent pushing bad descriptions**
- Refactored ability ui parser to use class attributes instead of passing around the ability key/data
- Fixed some issues with item localization
- Fixed S3 script


src/parser/maps.py
Surxe, line 1: Curious as to what allowed for removing all these remaps, more specifically, could you summarize where you find the data necessary to accomplish this? Maybe it could be applied similarly for other things, such as a couple attribute remaps.

Surxe (approved):
Love it!

src/parser/maps.py
hassayag: [This block ](https://github.com/deadlock-wiki/deadbot/pull/78/files#diff-4191ffc7130393ba87d55f5e1a6d4d03bd2035ac32772d9852732daf438fc607R459-R461)is the crux. The raw files stores a `m_strLocTokenOverride` for some data, which seems to be for this exact purpose. I mean they have to have them somewhere right? I guess I just didn't look hard enough previously

hassayag: Hopefully this is possible for other areas, cause manual overrides can become a little painful

Surxe: Good to know! Sadly, not really seeing any similar keys to help prevent some hero attribute remaps, but luckily theres very few.

Issue #79, opened 2024-10-16
Change BulletResist to BulletArmorDamageReduction scaling to detect it


Surxe:
Change BulletResist here
https://github.com/deadlock-wiki/deadbot/blob/15ebf8089bf5ab360ab921f6972ee13a8e9abee8/src/parser/maps.py#L116C51-L116C63

to BulletArmorDamageReduction such that it can be properly detected as scaling for Bebop. Currently, data shows he has 10% BulletArmorDamageReduction, and 1.25% BulletResist level scaling, meaning its not being detected as the same stat on deadlocked.wiki/Hero Comparison.

PR #80, opened 2024-10-16
Bugfix/bullet resist lvl scaling

Surxe:
Addresses #79 

BulletResist scaling was being considered a separate stat to base BulletResist, and was giving misleading info on deadlocked.wiki/Hero Comparison. This simply merges the stats.

Opposite to the proposal in the issue, I changed BulletArmorDamageReduction to BulletResist via hero attribute remap, rather than remapping BulletResist level scaling to BulletArmorDamageReduction.

Did the same for TechArmorDamageReduction to TechResist.

Uploaded hero and attribute data and confirmed its correct on deadlocked.wiki/Hero Comparison.


B13rg (approved):

Surxe:
I presume you wish to merge in the previous 2 minor patches before this gets merged?

hassayag (changes requested):
`BonusAttackRange` also has no base value of the same name. Does that also need a fix?

Surxe:
> `BonusAttackRange` also has no base value of the same name. Does that also need a fix?

Bebop scales attack range so its included. The get_meaningful_stats includes any stats that are scaled

hassayag:
> > `BonusAttackRange` also has no base value of the same name. Does that also need a fix?
> 
> Bebop scales attack range so its included. The get_meaningful_stats includes any stats that are scaled

Sure, but if you look at Bebop's hero data, there is no base value for `BonusAttackRange`. So isn't that the same issue as what you have with `BulletResist`?

Surxe:
> > > `BonusAttackRange` also has no base value of the same name. Does that also need a fix?
> > 
> > 
> > Bebop scales attack range so its included. The get_meaningful_stats includes any stats that are scaled
> 
> Sure, but if you look at Bebop's hero data, there is no base value for `BonusAttackRange`. So isn't that the same issue as what you have with `BulletResist`?

Ah I see. Luckily that's not an issue:
image

Its only an issue if the base stat doesnt match the scaling stat. There can be a scaling stat without a base stat, and vice versa.

hassayag (approved):
Sweet, off it can go 🚀

Surxe:
@hassayag Do you want to merge in the last 2 patches with a separate PR, or fine to merge them in with this?

hassayag:
> @hassayag Do you want to merge in the last 2 patches with a separate PR, or fine to merge them in with this?

Oh there was a patch yesterday, that's the only missing one i think. Let me merge that into develop myself first to keep the commit history clean

PR #81, opened 2024-10-16
Bugfix/Burst-dps-calc

Surxe:
Weapons that utilize burst (separate to bullets per shot) now calculate correctly.

Would greatly appreciate glossing over the new calc function. The extent to which I can get this fact-checked is quite limited, unfortunately.

https://github.com/deadlock-wiki/deadbot/blob/374cbf8f1571414ee9a6acf2771ebbe8367d03ed/src/parser/parsers/heroes.py#L233-L275
*edit, removed the debug print


Surxe:
This is ready to be merged.

B13rg (approved):
Looks great!
Nice job getting that calc figured out, that looks difficult!

PR #82, opened 2024-10-21
Feature/pull game files from s3

hassayag:
- **Created `DataTransfer` class to handle read/write of the S3 bucket**
  - `S3` class is now more general purpose
- Fixed directory of data uploaded to S3
- New `import_file` arg to trigger import at the beginning of deadbot process
- New `build_num` arg to select version to import. Default will get the latest version


Surxe (commented):
Did a very light skim and it looks good, though its well out of my wheelhouse so trust yourself to self-review and self-approve this one

B13rg (changes requested):
I like how each version has it's own folder; good defaults too

Be sure to add shutil to the poetry config with `poetry add shutil` as well

Looks really good, that's all that's needed to me!

hassayag:
> Be sure to add shutil to the poetry config with `poetry add shutil` as well

I didn't need to install shutil as it looks like its just a part of python by default

B13rg (approved):

B13rg:
Looks great!  Nice job!

PR #84, opened 2024-10-23
Feature/changelog tags & versioning prep

Surxe:
**Main changes**
- Changelog output now supports tagging lines with relevant tags. That line will appear on each wiki page corresponding to each tag. For example, "Abrams' Siphon Life now deals 5 more damage", would automatically have tags `Heroes`, `Abrams`, `Siphon Life`, and `Abilities`. This would cause the line to appear on wiki pages: /Heroes#Update history, /Abrams#Update History, /Siphon Life#Update history, and /Abilities#Update history. We can add support for more tags from their headers in `heading_to_tag`, or parse a specific tag from relevant data such as weapon items being assigned "Weapon Items" tag, or add them manually, lua modules and deadbot uploader is planned to do the rest.
- Changelogs now saved in the repository by their Valve-assigned build # (from the patch note url) instead of the date.
- The date and link for each valve-assigned build are also now parsed to `output-data/changelogs/changelogs.json`. This in preparation for a future PR that will focus on saving changelogs in deadbot by their Steam assigned manifest id. See the `#Get old version numbers` thread in the community discord on this process thus far.

**Other changes**
- hyphen prefixes replaced with asterisks for html bullet point integration
- "<tag>: " appearing in a line is no longer removed, as sometimes it removes needed context & its safer to leave each line as close to the original as possible

The {{PageRef}} has an initial version but is still very subject to change as it is planned to prioritize svg icons that are both light and dark theme compatible, which is waiting on dark theme release.

I added complete documentation on how I expect the Changelog editing process (after uploader) to go at [[Module:Changelog/doc]], as well as parsing for the list of Changelog Date pages displayed on [[Changelog Dates]], and some initial (front-facing) preparations at [[Changelog Tags]]


Surxe:
Ready for review. There are 300+ files in the filediff because of the storage change. Instead, to see the changes, run main.sh with Changelog on, then view `output-data/changelogs` directory, `fetch_changelogs`. and `parse_changelogs`

src/changelogs/fetch_changelogs.py
```diff
@@ -4,27 +4,55 @@
...
+        # write back
+
+        # this file is not overwritten even when update_existing is True
+        # many entries were manually added due to only the first page on the site has rss feed
+        changelogs_path = output_dir + '/changelogs.json'
+        existing_changelogs = json_utils.read(changelogs_path)
+        existing_changelogs.update(self.changelog_date_links)
+
+        # Sort existing_changelogs by key but numerically
+        keys = list(existing_changelogs.keys())
```
hassayag, line 49: what does this comment mean? Its not being sorted at this point

src/changelogs/fetch_changelogs.py
```diff
@@ -4,27 +4,55 @@
...
+        # Sort the keys numerically, not lexiconically
```
hassayag, line 50: Lexiconically??

This comment can just go, `.sort()` is pretty self explanitory

src/changelogs/fetch_changelogs.py
```diff
@@ -43,15 +71,34 @@ def _fetch_forum_changelogs(self, update_existing=False):
...
+            if not update_existing and (version in self.changelogs_by_version.keys()):
                 skip_num += 1
                 continue
             try:
                 full_text = '\n---\n'.join(self._fetch_update_html(entry.link))
             except Exception:
                 print(f'Issue with parsing RSS feed item {entry.link}')
-            self.changelogs_by_date[date] = full_text
+            if version is None or version == '':
+                print(date, entry.link)
```
hassayag, line 97: was this just for testing?

output-data/changelogs/versions/9652.json
hassayag, line 1: this changelog needs fixing up, not sure when or how it became empty

src/changelogs/parse_changelogs.py
```diff
@@ -36,56 +34,166 @@ def run(self, version, logs):
...
-                        icon = f'{{AbilityIcon|Ability={resource_name}|Size=20}}'
-                        new_log = log.replace(resource_name, f'{icon} {resource_name}')
-                        changelog[header][group_name][index] = new_log
+        new_changelog = changelog.copy()
+        for index, log in enumerate(changelog):
+            tags = log['Tags']
+            description = log['Description']
+            remaining_description = description
+            for tag in tags:
+                template = 'PageRef'
```
hassayag, line 160: With this pageref system, how does the wiki side figure out what, for example, what the link is for "Siphon Life"? Does it search through every hero?

src/changelogs/parse_changelogs.py
```diff
@@ -105,61 +213,21 @@ def _get_resources(self):
         resources.update(items)
         resources.update(abilities)
 
+        self.heroes = heroes
+        self.items = items
+        self.abilities = abilities
```
hassayag, line 218: `self.items` and `self.abilities` are unusued

output-data/changelogs/versions/38925.json
```diff
@@ -0,0 +1,67 @@
...
+        ]
+    },
+    {
+        "Description": "* T3 Golden Statues spawn time moved from 25 min to 30 min",
+        "Tags": [
+            "Other"
+        ]
+    },
+    {
+        "Description": "* Range to attack lane Guardians increased from 19m to 20m",
```
hassayag, line 39: we need to make considerations for neutrals/npcs as well. I'm happy to defer that to another PR though

hassayag (changes requested):

src/changelogs/parse_changelogs.py
```diff
@@ -36,56 +34,166 @@ def run(self, version, logs):
...
+            description = log['Description']
+            remaining_description = description
+            for tag in tags:
+                template = 'PageRef'
+
+                icon = '{{' + template + '|' + tag + '}}'
+
+                if tag in remaining_description:
+                    description = description.replace(tag, icon)
+                    new_changelog[index]['Description'] = description
```
hassayag, line 166: will the final description formatting be handled in another PR? Eg. arrows, bolding etc.

hassayag (changes requested):
Really great work! 
How we plan on deploying this? As we will have to make sure any existing contributions aren't just paved over (assuming they aren't handled by the parsed logs)

src/changelogs/fetch_changelogs.py
```diff
@@ -4,27 +4,55 @@
...
+from utils import json_utils
 
 
 class ChangelogFetcher:
     def __init__(self):
-        self.changelogs_by_date = {}
+        self.changelogs_by_version = {}
+        # layer1: version # as assigned by developers
+        # layer2: content (reworked in another pr, not gonna bother writing temporary
+        # stuff here for layer2-4)
```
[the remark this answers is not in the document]
Surxe: Whoops, this was referring to another PR, but I merged the content and closed the previous one. Removed the comments and using TypedDict now, which definitely looks a lot better now that I know the syntax!

src/changelogs/fetch_changelogs.py
Surxe: I think its worthwhile, though I meant lexicographically. 

Lexicographical sort:
1, 10, 100, 11, 12

Numerical sort:
1, 2, 3, ... 10, 11

src/changelogs/fetch_changelogs.py
Surxe: whoops, yes

output-data/changelogs/versions/9652.json
Surxe: looks like it was blank in `input-data` as its an old patch. Pasted it in and its all working now, nice catch!

src/changelogs/parse_changelogs.py
Surxe: Its still a bit up in the air right now, but ideally it doesn't need to know what type of resource it is. It just defaults to creating a link to [[Siphon Life]] and uses icon [[File:Siphon Life.svg]] or [[File:Siphon Life.png]]. This reduces the character count for each call drastically, which is important because editing changelogs on the frontend needs to be as streamlined and efficient as possible. Having only 1 parameter is the best way to accomplish that imo.

For now there's some logic on [[Module:PageRef]] that searches hero names and appends _MM to the filename, but this may get dropped down the line. I am planning on a PR designed for parsing icons from the files, so it will be revisited at that time.

output-data/changelogs/versions/38925.json
Surxe: Yeah we will likely find more tags we can parse for, though before I forgot I did these just now.

src/changelogs/parse_changelogs.py
Surxe: By arrows, do you mean like "Abrams health 600 --> 500"? I was planning on leaving them as is, but I suppose we could create something like {{Arrow|Right}} that displays a pretty arrow instead? 

If the page name appears in the changelog, itll automatically be bolded. Though, you just reminded me that I still need to make the headings appear in a wiki friendly format.

What other formatting do you have in mind?

Surxe:
> Really great work! How we plan on deploying this? As we will have to make sure any existing contributions aren't just paved over (assuming they aren't handled by the parsed logs)

Great question! Essentially, I am just going to need to spend a weekend merging the parsed data from here and the special formats that are done from the existing update pages. It will mostly be stuff like Shadow Transformation being nicknamed to Shadow Explosion as Valve renamed it. I have the main cases that require human-intervention documented on [[Module:Changelogs]] too!

src/changelogs/parse_changelogs.py
hassayag: The arrow thing was to match the wiki, but looking again, looks like they just use "to" now. Similar thing on other wikis, so maybe we don't need that 🤔

So maybe it doesn't actually need anything more, but I thought I'd open up the convo on it, as given the potential difficulty with deploying, we should make certain its as good as possible

output-data/changelogs/tag_groups.json
```diff
@@ -0,0 +1,22 @@
+[
+    "Abilities",
+    "Base Defenses",
+    "Creep",
+    "Guardian",
+    "Heroes",
+    "Items",
+    "Map",
+    "Mid-Boss",
```
hassayag, line 9: I think this needs checks for "Mid Boss"

src/changelogs/parse_changelogs.py
hassayag: But c reating a link won't link it to the page name, ie. `[[Siphon Life]]` needs to link to `Abrams/SIphon Life`

I'd suggest having an identifier, eg. `Item#Basic Mag`, `Ability#Siphon Life`. This removes the potential for conflicts, and helps narrow the search scope

src/changelogs/fetch_changelogs.py
hassayag: damn that's a new word to add to my.... lexicon

src/changelogs/parse_changelogs.py
Surxe: It will go directly to [[Siphon Life]] where the individual ability page will be, rather than the hero page [[Abrams#thing to Siphon Life]].

We could consider an identifier anyways down the line, though considering their wiki links won't be identifier-specific, it seems moot. For example, the earlier discussion on [[Spirit Lifesteal]] the stat vs the item. My ideal world would be to have [[Item/Spirit Lifesteal]] and [[Stat/Spirit Lifesteal]] instead, following OOP principles. Though, I know it seems too daunting to most contributors and would probably be opposed, so I don't think its worth the effort. It would also require revisiting the OOP restructure for the backend, not just the frontend.

src/changelogs/parse_changelogs.py
Surxe: Agreed. I'm sure that as I take a light skim through merging with existing changelogs, I'll find more tags that can be created before doing the actual merge. Though, that will still be many PR's down the line, as after this we need to hook it  up to the manifest-id-versioning branch.

hassayag:
Closed so i can reopen and trigger the deadlock-data PR 🤞

Surxe:
Summary of changes since last review:
* Added tons more tags, frontend-friendly list visible at [[Template:Changelogs Navbox]] (formatting is placeholder)
* Extracted lots of code within `parse_changelogs.run()` to separate functions for readability sake. Elaborated on their uses in depth in comments. 
* Remap-source is now used as `alt_name` parameter for `{{PageRef}}`, allowing for more embedments and removing the need to adjust for case sensitivity/plural. Example: "Fixed Rejuv buff" -> "Fixed {{PageRef|Rejuvenator|alt_name=Rejuv}} buff"
* Added `self.tag_parents` for assigning a tag's many **child <-- parent** relationship's in a more streamlined fashion. Detecting these relationships for say, resources such as **Siphon Life <-- Abrams**, still requires parsing however. Though, the relationships between group tags is verbosely defined with this map.
* Moved all maps/lists for tags to the top of the parser for easier reference, in order of their application. 
* Lines that are headings are now parsed adequately as having no tags, making them appear only on an update page i.e. [[Update:05-03-2024]] (fake example), and not a tag page's update section.
* Added hero lab's changelog in a similar format to `./input-data` and `/output-data/changelogs/changelogs.json` manually. Added some comments to `fetch_changelogs` for this process, but we need a more streamlined way to communicate this requirement documentation wise. Any ideas?

Surxe:
Today's changes
* 'Objective` tag used instead of `Base Defenses`, removing Trooper from the group
* Fixed edge case for nested PageRef calls
* All plural tags are now singular tags for consistency

src/changelogs/fetch_changelogs.py
```diff
@@ -5,53 +5,77 @@
...
-        # layer2: date
+    # Hero lab changelogs need to be added manually to
+    # ./input-data/raw-changelogs following the
+    # naming convention 'herolab_2024_10_29.txt'
+    # see the referenced file for example formatting
+
+    # Then they need to be added to
+    # /output-data/changelogs/changelogs.json following
+    # the same naming convention and formatted the same
+    # as other changelogs, but with "forum_id": null
```
hassayag, line 34: I'm not fond of TODOs in the code. Just put all this info in a git issue

src/changelogs/parse_changelogs.py
hassayag: Oh I thought we were doing ability pages like `<heroname>/<abilityname>`. If not, then that should in theory be okay.

While rare, I think naming conflicts does have potential to happen since the game is only going to get bigger. Eg. if they name an ability the same as an item, what would your solution be?

.gitignore
```diff
@@ -8,4 +8,5 @@ deadbot.egg-info
 .DS_Store
 .env
 docker-compose-local.yml
-.vscode
\ No newline at end of file
+.vscode
+/output-data/DepotDownloader/.DepotDownloader
```
hassayag, line 12: this is best included when the steam-related scripts are added

output-data/changelogs/raw/35440_herolab.txt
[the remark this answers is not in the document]
Surxe: Made up number for sorting purposes, named after previous build. The filename isn’t able to match the build # anymore unfortunately, hence the extra key for build # added to each in changelogs.json

src/changelogs/fetch_changelogs.py
Surxe: This is instructions for adding each hero lab changelog whenever they are released, so not a TODO perse. See my question in the last sentence of my summary on the 29th. Still hopeful for better ideas

Surxe:
@hassayag it won’t let me reply to your comment at https://github.com/deadlock-wiki/deadbot/pull/84#discussion_r1825075373 so I’ll reply here

The optimal (but would never be agreed upon) route is to have ability pages at /Ability/<abilityname>, hero pages at /Hero/<heroname>, etc. Instead, when we spot dupes during the uploader script, we will have it create the pages as /<abilityname> (Ability) and /<itemname> (Item). Im actually leaning towards no parentheses simply for better localization support.

src/changelogs/parse_changelogs.py
```diff
@@ -11,16 +11,168 @@ def __init__(self, output_dir):
...
+            'Weakened Patron': ['Patron', 'Objective'],
+            'Shrine': ['Objective'],
+            'Mid-Boss': ['NPC'],
+            'Weapon Item': ['Item'],
+            'Vitality Item': ['Item'],
+            'Spirit Item': ['Item'],
+            'Ability': ['Hero'],
+            'Soul Orb': ['Souls'],
+            'Urn': ['Souls'],
+        }
```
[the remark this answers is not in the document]
Surxe: good idea, done!

output-data/changelogs/raw/35440_herolab.txt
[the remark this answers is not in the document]
hassayag: There is `herolab_2024_10_29`. Should that also have a made up forum id?

src/changelogs/fetch_changelogs.py
hassayag: Fair the first paragraph seems helpful. Maybe that could go into a README for `input-data`? As that's where someone will be when adding the hero labs changelog

On the second paragraph, it sounds like you're informing the dev to add something to output-data? Nothing should be manually modified in output-data, as it will just be overwritten by the parser

src/changelogs/fetch_changelogs.py
```diff
@@ -4,27 +4,79 @@
...
-            os.makedirs(output_dir, exist_ok=True)
-            with open(output_dir + f'/{date}.txt', 'w', encoding='utf8') as f_out:
+        # Write raw changelog lines to files
+        for version, changelog in self.changelog_lines.items():
+            raw_output_dir = os.path.join(output_dir, 'raw')
+            os.makedirs(raw_output_dir, exist_ok=True)
+            with open(raw_output_dir + f'/{version}.txt', 'w', encoding='utf8') as f_out:
                 f_out.write(changelog)
 
+        # Write non-line data for the changelogs to 1 file
```
hassayag, line 58: what is non-line data?

src/changelogs/fetch_changelogs.py
```diff
@@ -4,27 +4,79 @@
...
+
+        # this file is not overwritten even when update_existing is True
+        # many entries were manually added due to only the first page on the site has rss feed
+        # herolab entries also need to be added manually as they are not on the forum at all
+
```
hassayag, line 63: Struggling to see how this comment block pertains to this function

src/changelogs/tags.py
```diff
@@ -0,0 +1,156 @@
...
+            'orbs',
+            'Orb',
+            'Orbs',
+            'soul',
+            'souls',
+            'rope',
+            'ropes',
+            'zipline',
+            'ziplines',
+        ]
```
hassayag, line 69: Can we just have a check for lowercase. ie. `input_str.lowercase() === 'orbs'`? Similar with the other tag lists

src/changelogs/tags.py
```diff
@@ -0,0 +1,156 @@
...
+            'Heroes': 'Hero',
+            'Abilities': 'Ability',
+            'Item Gameplay': 'Item',
+            'New Items': 'Item',
+            'Items': 'Item',
+            'Weapon Items': 'Weapon Item',
+            'Spirit Items': 'Spirit Item',
+            'Vitality Items': 'Vitality Item',
+            'Misc Gameplay': self.default_tag,
+            'Misc Gamepla': self.default_tag,
```
hassayag, line 94: gamepla 😂 
bro we need a LLM to make sure we get all of these

output-data/changelogs/raw/35440_herolab.txt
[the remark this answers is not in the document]
Surxe: Whoops, I see the confusion and forgot how I changed it. I renamed the file in output instead of input, leaving some duplicate. `herolab_2024_10_29` is the new format, not `<forum_id>_herolab`.

src/changelogs/fetch_changelogs.py
Surxe: The answer was so obvious, I should've realized thats why it felt so unintuitive lol. Moved that to input-data and added readme for input-data that has for now the changelog information.

src/changelogs/fetch_changelogs.py
Surxe: elaborated in comments

src/changelogs/fetch_changelogs.py
Surxe: removed the block and distributed parts of the comments to their relevant sections

src/changelogs/tags.py
Surxe: Could you elaborate on this idea? I just modified some of the guide phrasing in `tags.py`, but it doesn't need to strictly be lowercase, lowercase/shorter strings/single-words are more likely to be accidentally assigned when its actually referring to something else.

Surxe:
**Summary of changes since last review:**

1. More tags added
* Base Guardian
* Sapphire Hand / Amber Hand
* Cosmic Veil
* Crate
* Powerup
* Breakable (now parent to many previously added tags too)
* Sandbox
* Shop
* Minimap (merged into Map tag)
* Parry
* Light Melee, Heavy Melee, Melee
* Idol (merged into Urn)
* Pause
* Sinner's Sacrifice (Vault)
* Urn tag renamed to Soul Urn
* Flex Slot
* Hero lab related changes now use tag "HeroLab <heroname>"

2. Hero lab entries now added manually via `input-data` rather than `output-data`. Created a readme for this process as well, as it will be manual unless Valve uploads hero lab notes to the forums, or some image recognition is applied to read the ones from in game.


3. Tag child/parent hierarchy scripted into dictionary format at `tag_tree.json`, uploaded to [[Data:Changelog Tag Tree.json]], and used by [[Module:Changelog]] for use on [[Template:Changelogs Navbox]] (though admittedly its not actually a navbox yet, maybe som1 will have ideas for it down the line)

output-data/changelogs/tag_tree.json
```diff
@@ -0,0 +1,72 @@
...
+        "Shrine": {}
+    },
+    "Patron": {
+        "Weakened Patron": {}
+    },
+    "NPC": {
+        "Denizen": {},
+        "Creep": {
+            "Denizen": {},
+            "Trooper": {}
```
hassayag, line 45: trooper appears in three places. There should be a clear hierarchy

src/changelogs/fetch_changelogs.py
```diff
@@ -68,6 +77,10 @@ def changelogs_to_file(self, output_dir):
         # add any keys that are not yet present or have differing values,
         existing_changelogs.update(self.changelogs)
 
+        # add ones from input-data's changelogs.json, which currently include hero lab changelogs
+        input_changelogs = json_utils.read(os.path.join(input_dir, 'changelogs.json'))
+        existing_changelogs.update(input_changelogs)
```
hassayag, line 82: I'm confused what `changelogs.json` is doing. It only contains a single entry for hero-lab. Is there a problem with just reading it like any other raw-changelog file?

src/parser/parsers/ability_ui.py
```diff
@@ -45,6 +45,12 @@ def run(self):
                     if parsed_ui is not None:
                         hero_abilities[self.ability_index] = parsed_ui
                 except Exception as e:
+                    # TEMPORARY, IDK HOW TO GET AROUND IT PERMANENTLY
+                    if ability['Key'] in [
+                        'ability_nano_proximity_ritual',
+                        'ability_viper_snakedash',
+                    ]:
+                        continue
```
hassayag, line 53: I don't see an error without this, maybe an update has changed it

output-data/changelogs/tag_tree.json
Surxe: Looks like there is two separate issues here:

The first being that Trooper is under NPC when it only needs to be under Creep (a child of NPC).

The second being that:
```
"Creep": {
        "Denizen": {},
        "Trooper": {}
}
 ```
Is appearing both at the root level, and under NPC. So I'll need to look into fixing that, nice catch!

File: output-data/changelogs/tag_groups.json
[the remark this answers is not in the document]
Comment (Surxe) [reply output-data/changelogs/tag_groups.json:1]: It was used before `tag_tree` to help communicate warnings for changes in parsed tags, it *should* be unneeded now so I'll refactor around that, good find!

I can't come up with a better terminology but, group tags refers to tags that aren't individual Hero names, Item names, or Ability names. This is determined by the `is_group_tag` parameter in `register_tag()`. It's purpose is simply to prevent these 100+ tags from bloating the displayed hierarchy into a state of being unreadable. Also, not finding where I wrote the comments on this definition, so guess I'll add this for it lol

File: src/changelogs/fetch_changelogs.py (hunk 18 above)
Comment (Surxe) [reply src/changelogs/fetch_changelogs.py:82]: This herolab entry can't be read like other raw-changelog files because it doesn't have a forum post to read off of, meaning it has no assigned date. After this is merged and we look at integrating it with the manifest-id versioning branch, a cleaner route may present itself, though I doubt it. As then, we won't be trying to assign it a date, but rather assign it a manifest-id or a ClientVersion or something else to attach it to. I'm unsure, but for now the date for herolab changelogs would need to be added this way.

File: output-data/changelogs/tag_groups.json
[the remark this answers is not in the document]
Comment (hassayag) [reply output-data/changelogs/tag_groups.json:1]: > it should be unneeded now

Am i reading this right, so if it's not needed, why can't we delete it?

File: src/changelogs/fetch_changelogs.py (hunk 18 above)
Comment (hassayag) [reply src/changelogs/fetch_changelogs.py:82]: Oh the way the git diff showed me the file made it look like there was only a single entry in it, but actually every changelog is represented there, so that's all good

File: output-data/changelogs/tag_groups.json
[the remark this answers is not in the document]
Comment (Surxe) [reply output-data/changelogs/tag_groups.json:1]: Sorry for the confusion, it will be deleted, I just meant that I also need to remove the internal diff-check that runs off of it.

File: src/changelogs/parse_changelogs.py (hunk 4 above)
Comment (hassayag) [reply src/changelogs/parse_changelogs.py:206]: @Surxe comment still outstanding. did we discuss this in a chat? Can't remember

File: src/changelogs/tags.py [hunk 20]
```diff
@@ -275,3 +274,102 @@ def __init__(self, default_tag):
             'HeroLab Magician': ['Magician', 'Hero'],
             'HeroLab Viper': ['Viper', 'Hero'],
         }
```
hassayag, line 276: It seems like we have two places that define tag relationships. Shouldn't a simple tag tree, as you have below, cover that fully?

Surxe: Correct, newly added self.tag_tree is going to replace self.parents. I haven't finished quite yet so I haven't removed the old one, still a few things to look over.

src/changelogs/parse_changelogs.py
Surxe: > @Surxe comment still outstanding. did we discuss this in a chat? Can't remember

Was buggin out on mobile earlier, wouldn't let me reply directly. You can see my reply here https://github.com/deadlock-wiki/deadbot/pull/84#issuecomment-2450861376

Surxe:
**Changes since last review:**
* `self.parents` being defined then transformed into `self.tag_tree` is removed. Now, `self.tag_tree` is defined, then transformed into `self.parent_lookup` (structure identical to the old `self.parents`). 
* Since this format now more directly corresponds to the outputted `tag_tree.json` dict (as opposed to the old `tag_groups.json` list), group tags are no longer needed. As such, everything related to group tags they were completely removed, greatly simplifying the comments needed to explain several processes.
* Removed the need for HeroLab related tags to be specified in `self.tag_tree` manually
* Added indicators to `tag_tree.json` where instance tags would appear (like Abrams, Basic Magazine, etc.), via `<Hero Name>`, `<Ability Name>`, etc. 
* Fixed tags that were accidentally in the tag_tree in two different locations. They now appear in only the furthest depth. 
* `self.tag_tree` and therefore `tag_tree.json` now use a custom order that isn't restricted to alphabetical like before.
* Today's HeroLabs patch added to `input-data`

**Specific tag related changes:**
* Guardian, Base Guardian, and Walker tags are now only children to Objective, and not NPC. This is not only because they are arguably more robot than an NPC, but also to reinforce the ideal state of tags only appearing in the tag_tree once.
* Parry tag is now a child to the Melee tag
* Gun tag added, and as a child to the Hero tag

src/changelogs/tags.py
hassayag: I'm suggesting you remove any checks for casing by treating upper and lower case the same. Eg. if you check just for lower case, you don't need to look for `pits` and `Pits`. If you can point out major issues with this then fair enough

input-data/README.md
```diff
@@ -0,0 +1,13 @@
+# Adding input-data
+
+# Changelogs
+Historical changelogs:
```
hassayag, line 4: Improve markdown heading hierarchy. ie. use `## Changelogs`, `### Historical changelogs`

input-data/README.md
```diff
@@ -0,0 +1,13 @@
...
+The RSS feed that stores previous forum changelog posts <https://forums.playdeadlock.com/forums/changelog.10/> only stores the latest few months of changelogs. As such, older changelogs have been added here manually already.
+
+Herolab changelogs:
+Hero lab changelogs do not exist on the forum, and only exist in game. The patch notes for it need to be created in its own file in `input-data/raw-changelogs/here.txt`, with the file name formatted as `herolab_yyyy_mm_dd.txt`. See the content of `input-data/raw-changelogs/herolab_2024_10_29.txt` for an example.
+
+Add the configuration for the hero lab changelog to `input-data/changelogs.json`
+
+# Note
+Note that the paths involving `input-data` are assuming the input-data path's parameter is defaulted.
```
hassayag, line 13: What is the input data path parameter? Unsure what this means

src/changelogs/parse_changelogs.py
```diff
@@ -11,16 +11,168 @@ def __init__(self, output_dir):
...
+        # list of all unique tags excluding resource tags
+        # since they exclude resource tags (instances), they are referred
+        # to as "group tags" / "tag groups"
         self.unique_tag_groups = [self.default_tag]
 
+        # Tags to register if they are found in the changelog line
+        # match by text
+        # never put lowercase tags here, as they are likely to be
+        # part of a longer string. Put them in self.tags_match_word
+        # instead
```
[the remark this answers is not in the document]
hassayag: > # "Return Fire changed from...", "urn" would be incorrectly tagged

I'm not sure if this is what you added for this comment, but either way, every string check should be for the whole word. No partial matches. Ie. just make sure it has a space either side (depending on the position of the word)

Unless I'm missing something, I don't see why we need separate lists for `match_word` and `match_text`

src/changelogs/tags.py
```diff
@@ -0,0 +1,342 @@
...
+            'Amber Hand': {},
+            'Sandbox': {},
+            'Pause': {},
+            'Bounce Pad': {},
+            'Rope': {},
+            'Flex Slot': {},
+            'Greenwich': {},
+            'York Avenue': {},
+            'Broadway': {},
+            'Park Avenue': {},
```
hassayag, line 319: should the lane names be here?

src/changelogs/parse_changelogs.py
```diff
@@ -93,73 +235,54 @@ def _get_resources(self):
...
+
+    def _write_tag_tree(self, tag_tree):
+        # Add <hero>, <ability>, <item> etc. to tag tree
+        # to show where instance tags would appear (such as Abrams, Basic Magazine, Siphon Life)
+        tag_tree['Hero']['<Hero Name>'] = {}
+        tag_tree['Hero']['<HeroLab Hero Name>'] = {'HeroLab <HeroLab Hero Name>': {}}
+        tag_tree['Hero']['Ability']['<Ability Name>'] = {}
+        tag_tree['Item']['Weapon Item']['<Weapon Item Name>'] = {}
+        tag_tree['Item']['Vitality Item']['<Vitality Item Name>'] = {}
+        tag_tree['Item']['Spirit Item']['<Spirit Item Name>'] = {}
```
hassayag, line 273: Is anything stopping us from just including all of this in `tags.py` directly?

src/changelogs/tags.py
```diff
@@ -0,0 +1,342 @@
...
+                'Bridge': {},
+                'Hotel': {},
+                'Pharmacy': {},
+                'Penn Station': {},
+                'Grand Central': {},
+                'Pit': {},
+            },
+            'Powerup': {},
+            'Zipline': {},
+            'Rejuvenator': {},
```
hassayag, line 307: Would rejuvenator make sense in `mid-boss`?

src/changelogs/tags.py
```diff
@@ -0,0 +1,342 @@
...
+            'Cosmic Veil': {},
+            'Sapphire Hand': {},
+            'Amber Hand': {},
+            'Sandbox': {},
+            'Pause': {},
+            'Bounce Pad': {},
+            'Rope': {},
+            'Flex Slot': {},
```
hassayag, line 315: flex slot should be in `items`

src/changelogs/tags.py
```diff
@@ -0,0 +1,342 @@
...
+                'Penn Station': {},
+                'Grand Central': {},
+                'Pit': {},
+            },
+            'Powerup': {},
+            'Zipline': {},
+            'Rejuvenator': {},
+            'Cosmic Veil': {},
+            'Sapphire Hand': {},
+            'Amber Hand': {},
```
hassayag, line 310: Perhaps these fit inside a `Lore` group? idk enough about what those are

src/changelogs/tags.py
```diff
@@ -0,0 +1,342 @@
...
+        # also see self.parent_lookup
+        self.tag_tree = {
+            'Item': {'Weapon Item': {}, 'Vitality Item': {}, 'Spirit Item': {}},
+            'Hero': {'Ability': {}, 'Gun': {}},
+            'Melee': {
+                'Light Melee': {},
+                'Heavy Melee': {},
+                'Parry': {},
+            },
+            'NPC': {'Denizen': {}, 'Creep': {'Denizen': {}, 'Trooper': {}}, 'Mid-Boss': {}},
```
hassayag, line 282: duplicate `Denizen`

src/changelogs/tags.py
```diff
@@ -0,0 +1,342 @@
...
+
+        # Relations between a child and parent tag.
+        # i.e. Abrams is a parent to Siphon Life, and a child to Hero
+        # tags below are after _remap_tag() is called
+        # key = parent
+        # value = children
+        # also see self.parent_lookup
+        self.tag_tree = {
+            'Item': {'Weapon Item': {}, 'Vitality Item': {}, 'Spirit Item': {}},
+            'Hero': {'Ability': {}, 'Gun': {}},
```
hassayag, line 276: Gun is named "Weapon" on the wiki

src/changelogs/parse_changelogs.py
```diff
@@ -1,88 +1,230 @@
...
+        for resource_key in self.resources:
+            resource = self.resources[resource_key]
+            resource_type = resource['Type']
+            resource_name = resource['Name']
+
+            if resource_name is None:
+                continue
+
+            # Skip disabled items
+            if resource_type == 'Item' and resource.get('Disabled', False):
```
hassayag, line 82: Not sure about skipping these. eg. Soul Rebirth was recently removed, if we ignored disabled items, then we would have missed that tag

src/changelogs/tags.py
Surxe: Don't see a reason not to include em

src/changelogs/parse_changelogs.py
Surxe: Hadn't thought about it, you absolutely right. Moved _write_tag_tree to tags.py.

src/changelogs/tags.py
Surxe: I thought about this recently;

The idea is, whenever we find changelogs regarding Rejuvenator, do we want them to also end up on the Mid-Boss page. Or in other terms, when we visit the Mid-Boss#Update history section, do we want to see changes regarding Rejuvenator. Personally, I would say no. They may be in a similar mechanic but I'm not sure thats sufficient to nest Rejuv under Mid-Boss. 

The alternative I thought of was to instead nest both of these under some tag like "Mid-Boss Mechanics". But, I can't think of a better name and it seems a bit overkill. If you can think of a more appropriate name lmk though

src/changelogs/tags.py
Surxe: good catch

src/changelogs/tags.py
Surxe: I thought about this too, though looking at where they are used, it is very generic. i.e.
"* {{PageRef|Sapphire Hand|alt_name=Sapphire}} side materials tint color changed to white"

Here it's just being used to represent the team. My original idea was to nest them under something more generic like "Team", but it feels too generic to warrant it's existence. As I'm writing this I realize it does warrant it's existence lol. Moving em under "Team" tag.

src/changelogs/tags.py
Surxe: good catch

src/changelogs/parse_changelogs.py
Surxe: I would agree, I didn't have this initially. The reason I restricted it to items only is specifically because there are disabled items with what are I presume placeholder names like `Fire Rate` and `Health Regen`. This caused them to be flagged as Items in tons of places where it was obviously referencing a stat. 

Looking through item data now, it seems it's actually just these two. Instead, I'll replace this condition with just not being either of these two specific names.

Surxe:
Changes since last review:
* Improved the changelog readme
* `_write_tag_tree` moved to `tags.py`
* Removed the duplicate `Denizen` tag
* `Sapphire Hand` and `Amber Hand` now children to `Team` tag
* `Gun` tag renamed to `Weapon`
* Skip problematic disabled items rather than all disabled items
* **Check for resources in order of the length of their name. I.e. if the text is `Smoke Bomb`, the tag assigned will now reliably be `Smoke Bomb` instead of `Smoke` (a disabled item)**

src/changelogs/tags.py
hassayag: I totally see your point, I think its fine then. The wiki should allow for multiple filters, so you could in theory ask for both "Mid Boss" and "Rejuvinator" without the need of a parent tag

src/changelogs/tags.py
hassayag: Oh yeah those are the names of the teams 🤦. Yeah no harm in having it

src/changelogs/parse_changelogs.py
hassayag: Ohh interesting. Hopefully no more terrible item names like that in future

src/changelogs/tags.py
hassayag: Would it fit into the "Map" parent?

Surxe: I thought about this too; I added it then thought about adding Bounce Pads, Rope, Zipline, etc. But then I realized that if these are referenced in gameplay changes, then its nothing to do with the map, so I removed the lane names too. Though thinking again now, I think your right; the lane names can go into Map because they should only be referenced in map-related changes, similar to the other map locations.

src/changelogs/tags.py
hassayag: Still needs renaming on the tag tree

src/changelogs/tags.py
hassayag: Its definitely a difficult thing to figure out, since it depends how the wiki is structured, which can change a lot. 

Possibly we could consider excluding parent tags from the changelogs themselves, and have a lookup table to find any parents? Not one for this PR, but it would mean these decisions are less crucial, since a change to the tag tree wouldn't require reparsing the changelogs

hassayag: It also means tagging changelogs manually is simpler. Eg. if an item change is missing its tag, you don't need to add the item name, "Item", and "Weapon Item", you just need to add the item name

src/changelogs/tags.py
Surxe: That would be ideal yeah. There's certainly some pros to the way it is right now, but with the size the tags lists have gotten to recently, it probably warrants a change nowadays. We can amend this in another PR as you mentioned in discord.

src/changelogs/tags.py
hassayag: Not quite, it's more to do with removing any redundant tags from the changelog files. Eg. if a changelog contains "Weakened Patron", it wouldn't need to have "Patron" or "Objective" explicitly written in the tags list, as that can be defined by a child-parent map stored separate

hassayag: But its only a light suggestion, I'm not sure thats so great, as having them baked in keeps the wiki simpler

Surxe: Hmm, I *think* I understand the idea. But yeah this could be a route for improvement down the line. Luckily, after the `tags.py` is configured, there isn't much that needs to be maintained with it, as in games like this 90% of new content is new resources (i.e. new heroes, items), as opposed to new resource types or mechanics which would require more tags to be handled explicitly.

hassayag (approved):
Probably the largest diff in history, excellent work!

       ✨                        🌔            ✨
         👑         ✨
         🚀                                  ✨
 ✨

PR #85, opened 2024-10-24
Feature/GitHub action to parse game files

hassayag:
This PR is designed to eliminate all local parsing of game files, as that means we can ensure a more consistent and reliable environment when carrying out the crucial process of parsing and uploading game files.

The new workflow takes all output-data off of deadbot repo and stores it on deadlock-data repo. For now, I have left the original output-data so we test the workflow and double check its smoooth
## New Developer Workflow

The following flows rely on decompiled data on S3, which to start will involve me pushing new game files each patch
### On PR being opened 
    1. Working branch is created on deadlock-data repo
    2. PR is opened on deadlock-data repo, with a link back to the deadbot PR
    3. New PR is linked on the deadbot repo for easy viewing of diff

### Commits to working branch
    1. Decompiled data is pulled from s3 and parsed
    2. Parsed data is commited to new branch

### On workflow dispatch or merge into master:
    1. Decompiled data is pulled from s3 and parsed
    2. Parsed data is commited to respective branch, develop or master on deadlock-data repo
    3. Output data is pushed to the wiki

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/7)_


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+          git pull origin ${{ needs.get-branch.outputs.branch }} -X ours
+          git switch -c ${{ needs.get-branch.outputs.branch }}
+          git push --set-upstream origin ${{ needs.get-branch.outputs.branch }}
+
+      - name: Pull branch
+        working-directory: deadlock-data 
+        run: git pull origin ${{ needs.get-branch.outputs.branch }} -X ours     
+
+      - name: Copy files to data repo
+        run: cp -r ./deadbot/output-data ./deadlock-data/data/current
```
B13rg, line 92: the data repo has some scripts that can assist, you might be able to copy them and mold them to your needs: https://github.com/deadlock-wiki/deadlock-data/blob/develop/Taskfile.yml

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,127 @@
...
+    needs: get-branch
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          path: deadbot
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
B13rg, line 49: Not needed if using docker

hassayag:
Sorry this was meant to stay in draft oopsy. It is 90% there though, so comments are welcomed, just know it's not 100% working

src/utils/csv_writer.py
```diff
@@ -1,5 +1,8 @@
 import pandas as pd
 import os
+from dotenv import load_dotenv
+
+load_dotenv()
```
B13rg, line 5: Not needed

Surxe: We found it necessary when trying on another machine even with WSL installed. We couldn't figure out why it was needed, just that it was. If you know why though lmk too :)

B13rg (approved):
Only change is removing the dotenv from csv_writer, if it needs vars it can references the constants file

I was thinking that the deadbot source changes should be separate from the data-update workflow, and instead of building the docker container as part of the workflow, it should just pull it. But now that I think about it, that workflow would just live in the deadlock-data repo, separate from all this.  When the "source of truth" is moved to the deadlock-data repo, the automation can go along with it.

src/utils/csv_writer.py
B13rg: ahh the joys of the shell, no problem!

hassayag: I just tested it again without, turns out its fine, so not sure why I added it, or what changed since that commit. 

I will revert this, since in your case @Surxe, you can have a single `load_dotenv()` call in `deadbot.py` if thats the only solution

hassayag:
> Only change is removing the dotenv from csv_writer, if it needs vars it can references the constants file
> 
> I was thinking that the deadbot source changes should be separate from the data-update workflow, and instead of building the docker container as part of the workflow, it should just pull it. But now that I think about it, that workflow would just live in the deadlock-data repo, separate from all this. When the "source of truth" is moved to the deadlock-data repo, the automation can go along with it.

It has to build the docker container in this case, since it is using the working branch, not the released version

src/utils/csv_writer.py
Surxe: Sure thing!

PR #90, opened 2024-10-26
Feature/Soul-Unlocks-parse

Surxe:
Parses soul unlock related data.

Data is uploaded to https://deadlocked.wiki/Data:SoulUnlockData.json

Module is at https://deadlocked.wiki/Module:SoulUnlock

Module currently supports two structures, the one uploaded, and an aggregated structure. Has localization support too so it all works nice 'n dandy already :)

### DIFF
--- a/src/parser/maps.py
+++ b/src/parser/maps.py
@@ -21,7 +21,7 @@ def get_target_type(value):
 
 
 SOUL_UNLOCK_MAP = {
-    'm_unRequiredGold': 'RequiredGold',
+    'm_unRequiredGold': 'RequiredSouls',
     'EAbilityUnlocks': 'AbilityUnlocks',
     'EAbilityPoints': 'AbilityPoints',
     'm_bUseStandardUpgrade': 'PowerIncrease',
--- a/src/parser/maps.py
+++ b/src/parser/maps.py
@@ -19,6 +19,7 @@ def get_target_type(value):
         raise Exception(f'{value} is not a valid target type')
     return TARGET_TYPE_MAP.get(value)
 
+
 SOUL_UNLOCK_MAP = {
     'm_unRequiredGold': 'RequiredGold',
     'EAbilityUnlocks': 'AbilityUnlocks',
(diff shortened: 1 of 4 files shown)


src/parser/maps.py
```diff
@@ -20,6 +20,14 @@ def get_target_type(value):
     return TARGET_TYPE_MAP.get(value)
 
 
+SOUL_UNLOCK_MAP = {
+    'm_unRequiredGold': 'RequiredGold',
```
hassayag, line 24: Is there a good reason to not call this souls?

src/parser/parsers/souls.py
```diff
@@ -0,0 +1,78 @@
+from parser.maps import SOUL_UNLOCK_MAP
+
+
+class SoulUnlockParser:
```
hassayag, line 4: Brief description of what soul unlock means would be good

src/parser/parsers/souls.py
```diff
@@ -0,0 +1,78 @@
...
+                if parsed_soul_unlocks == {}:
+                    parsed_soul_unlocks = hero_soul_unlocks
+                    hero_key_of_parsed_soul_unlocks = hero_key
+                # Confirm that the soul unlocks is the same as the previous hero's soul unlocks
+                else:
+                    if parsed_soul_unlocks != hero_soul_unlocks:
+                        raise ValueError(
+                            'Soul unlocks do not match between heroes'
+                            + f'{hero_key_of_parsed_soul_unlocks} and {hero_key}'
+                        )
```
hassayag, line 28: ah very nice, i just had this thought pop up, since they show them per hero in game

hassayag (changes requested):
Very nice, future proofs us for inevitable changes

src/parser/maps.py
Surxe: Changed to RequiredSouls on backend and frontend

src/parser/parsers/souls.py
Surxe: Good point, added!

hassayag (approved):

PR #93, opened 2024-10-31
Feature/Verbose loguru and better warns/excepts

Surxe:
As the repository grows larger and larger, the number of print statements and warnings also grows, decreasing the readability of the terminal. This pull request focuses on adding the parameter `-v` / `--verbose` and `.env`'s `VERBOSE` to remove many prints. 

Additionally, there are some light fixes to deprecated functions/modules to remove the following warnings
```
F:\Python\Lib\site-packages\dateutil\tz\tz.py:37: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to reppresent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
  EPOCH = datetime.datetime.utcfromtimestamp(0)

F:\Repositories\deadbot\src\utils\csv_writer.py:8: FutureWarning: DataFrame.applymap has been deprecated. Use DataFrame.map instead.
  df = df.applymap(convert_array_to_string)
F:\Repositories\deadbot\src\utils\csv_writer.py:8: FutureWarning: DataFrame.applymap has been deprecated. Use DataFrame.map instead.
  df = df.applymap(convert_array_to_string)
```

*edit: the fix for the first one seems to have actually been in a hidden file, so ignore that. Just the fix for the bottom csv error is included in this PR.

With verbose disabled, a sample output would be:
image

With verbose enabled, a sample output is now simply:
image

The only prints that run unconditionally are the above which are in the `DeadBot.main()` method, as well as warnings.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/51)_


Surxe:
@B13rg Would you mind reviewing this one? There isn't any particular rush, but Saag will be out until next Tuesday, and I figure this is more your realm :)

hassayag:
Closing so i can reopen with new gha

hassayag:
Okay one more time 😓

hassayag:
This can be improved upon, as  `if self.VERBOSE` is too messy. Look into using something like https://github.com/Delgan/loguru and make use of log levels

Surxe:
> This can be improved upon, as `if self.VERBOSE` is too messy. Look into using something like https://github.com/Delgan/loguru and make use of log levels

This looks really neat, mind if I suggest we look into it later down the line though via another PR? In order to utilize levels, it looks like it will require each contributor to at minimum become familiar with its library. I don't think this warrants the (currently) minimal benefit it would yield. At least personally, `if self.VERBOSE` is more than sufficient for now, given it will of course provide an upgrade to the current system.

Down the line, it would require usage of the verbose parameter anyways, as we could use it to alter the log levels to a higher/lower level. Looks simple to integrate too!

hassayag:
> In order to utilize levels, it looks like it will require each contributor to at minimum become familiar with its library

I'm not sure what there is to be familiar with, its a pretty standard api

Surxe:
> > In order to utilize levels, it looks like it will require each contributor to at minimum become familiar with its library
> 
> I'm not sure what there is to be familiar with, its a pretty standard api

If you want it added directly into this PR, feel free to implement it, but it’s not something I’m going to be spending my time on currently. I think it’s more than reasonable to have that in a later PR.

hassayag:
> > > In order to utilize levels, it looks like it will require each contributor to at minimum become familiar with its library
> > 
> > 
> > I'm not sure what there is to be familiar with, its a pretty standard api
> 
> If you want it added directly into this PR, feel free to implement it, but it’s not something I’m going to be spending my time on currently. I think it’s more than reasonable to have that in a later PR.

Doing something once correctly is easier in my books. Measure twice cut once or whatever the analogy might be. If this feature was crucial I might be more lenient on that, but logging is so inconsequential

Surxe:
Id prefer to cut

> > > > In order to utilize levels, it looks like it will require each contributor to at minimum become familiar with its library
> > > 
> > > 
> > > I'm not sure what there is to be familiar with, its a pretty standard api
> > 
> > 
> > If you want it added directly into this PR, feel free to implement it, but it’s not something I’m going to be spending my time on currently. I think it’s more than reasonable to have that in a later PR.
> 
> Doing something once correctly is easier in my books. Measure twice cut once or whatever the analogy might be. If this feature was crucial I might be more lenient on that, but logging is so inconsequential

The first measurements have already been made and cut. There’s no benefit to not using it in place of the current system when its a pre-requisite of additional measurements anyways. If you want to make more measurements and cut it again down the line, feel free.

Surxe:
Had to add some extra logic to `parser.py`'s `_load_localizations` to avoid some duplicate keys from raising exceptions

hassayag:
I've added setup of logguru, i'll leave the rest to you

src/utils/parameters.py
```diff
@@ -48,6 +51,12 @@ def arg_group_base(parser):
         help='Build number of the game files to be used. Defaults to current build',
         default=os.getenv('BUILD_NUM', None),
     )
+    group_base.add_argument(
+        '-v',
+        '--verbose',
```
hassayag, line 56: i get an error when using `--verbose`

src/utils/parameters.py
```diff
@@ -1,5 +1,8 @@
 import os
 import argparse
+from dotenv import load_dotenv
+
+load_dotenv()
```
hassayag, line 5: are these necessary? the one in deadbot.py should load your env

src/utils/csv_writer.py
```diff
@@ -5,7 +5,7 @@
 def export_json_file_to_csv(file_name, OUTPUT_DIR):
     os.makedirs(f'{OUTPUT_DIR}/json', exist_ok=True)
     df = pd.read_json(f'{OUTPUT_DIR}/json/{file_name}.json').transpose()
-    df = df.applymap(convert_array_to_string)
+    df = df.map(convert_array_to_string)
```
hassayag, line 8: funnily enough mine only works with `applymap`. What pandas version are you on?

src/utils/parameters.py
Surxe: It seems I'm confused on how this works; I can't find any other instances of "build_num" or "BUILD_NUM" in the repo that cause it to work where "VERBOSE" doesn't. Do you know?

src/utils/parameters.py
Surxe: Isn't necessary on my desktop, but it might've been on my laptop. I can't remember, but your probably right it should only need the one in deadbot.py. I'll remove it from here for now and if issues arise I'll fix it elsewhere.

src/utils/csv_writer.py
Surxe: I just re-installed it to double check, the warning is ```
F:\Repositories\deadbot\src\utils\csv_writer.py:8: FutureWarning: DataFrame.applymap has been deprecated. Use DataFrame.map instead.
  df = df.applymap(convert_array_to_string)```
Can you double check your version is up to date? I would presume that if applymap is deprecated, that's the older one.

Surxe:
> I've added setup of logguru, i'll leave the rest to you

I'm confused on when to use logger.trace vs logger.info vs logger.warn. The one example that uses logger.warn so far doesn't read like its a warning

hassayag:
> > I've added setup of logguru, i'll leave the rest to you
> 
> I'm confused on when to use logger.trace vs logger.info vs logger.warn. The one example that uses logger.warn so far doesn't read like its a warning

Oh yeah I think I was just testing the log levels. That should be info

Surxe:
@hassayag Ok think I got it all working. Just need confirmation on the bit in `csv_writer`

src/utils/csv_writer.py
hassayag: It's probably cause the version in pyproject.toml is 2.0.1. Just use that version for now, we can upgrade it in another pr

Surxe: > It's probably cause the version in pyproject.toml is 2.0.1. Just use that version for now, we can upgrade it in another pr

How do I replace with an older version? My dependency installation process is still screwed up, needa use python3 -m for everything, so wanna make sure I get it right :/

hassayag: pip is still pip, whether it has python3 before or not. You should be able to find the syntax to pip install a specific version

Surxe: > pip is still pip, whether it has python3 before or not. You should be able to find the syntax to pip install a specific version

Fair enough, but the syntax isn't my problem. This PR can wait until after the pandas version is updated in the pyproject.tml.

hassayag: Well you did ask how to install it. Either way, this PR should have no changes related to pandas. Just remove the change and let the other issue handle it, meaning this won't be blocked.

Also needs updating with develop as there should be no output-data anymore

Surxe:
**New summary**
Adds the `VERBOSE` parameter, loguru logging, and update some exceptions/warnings now that output-data is no longer git-diffed.

- csv_writer's apply_map change was reverted, pyproject.toml can be updated to use latest pandas version in another PR.
- The "non-constant stats have changed" warning now lists added/removed keys
- Added fatal exception handling if decompilation cmd fails.
- More specific logic for checking for localization-dupes that may actually be problematic

Surxe:
@hassayag This PR is ready but getting a check error with "PR Synchronize" above, could you take a look when your free?

hassayag:
reopening for deadlock-data branch creation

src/parser/parser.py
```diff
@@ -92,13 +95,30 @@ def _merge_localizations(self, language, group, data):
...
+                    elif new_is_english and not old_is_english:
+                        pass  # keep the old localized string
+                    else:
+                        # If both values are english or neither are english, raise an exception
+                        logger.trace(
+                            f'Key {key} with value {value} already exists'
+                            + f'in {language} localization '
+                            + f'data with value {current_value}.'
+                            + f' English value: {english_value}'
+                        )
```
hassayag, line 118: What's the motivation around this?

src/parser/parser.py
```diff
@@ -129,14 +149,18 @@ def _parse_heroes(self, parsed_abilities):
...
+        )
+        if current_meaningful_stats != parsed_meaningful_stats:
+            current_keys = set(current_meaningful_stats.keys())
+            new_keys = set(parsed_meaningful_stats.keys())
+            logger.warning(
+                'Non-constant stats have changed. '
                 + "Please update [[Module:HeroData]]'s write_hero_comparison_table "
                 + 'lua function for the [[Hero Comparison]] page.'
+                + f'\nAdded keys: {new_keys - current_keys}'
+                + f'\nRemoved keys: {current_keys - new_keys}'
```
hassayag, line 163: With the removal of output-data, this is no longer possible. Instead you could add the wiki json file to your watch list to keep track of any changes

src/utils/parameters.py
hassayag: build num is a bad example to use, as that takes in an argument. You want one that doesn't, so take a look at some other existing options for comparison

src/parser/parser.py
Surxe: Previously this area threw a hard exception, this *very* old change made it a bit more accurate to the problem it was trying to detect. Now that its just a soft warning its not really necessary and overly specific, will remove.

src/parser/parser.py
Surxe: Hmm I guess that's true when ran via GHA unfortunately. We will want to look into replacements for this soon. Watch list is not a good system because wiki diffing data structures does not work. i.e. some attributes of hero_wraith changed on line70-72, but the line that says its wraith's attributes are on line60. Since the wiki diff will only show line70-72, we can't see what changed accurately.

Also, recall that this file is not uploaded to the wiki, its only purpose is to more accurately maintain what stats are variable and which ones are constants for all heroes. To fix deadbot not erroring, I added a small check to ensure the path exists before comparing. For now, we can keep this check even though it only works locally.

We can discuss ideas for the replacement later and elsewhere, but if you really want to leave out reading from the wiki from this repo then we can either build it into the deadlock-data repo or make a 3rd repo, as I have lots of ideas to improve the workflow that involve reading from the wiki.

src/parser/parser.py
Surxe: For example, `Key SpiritStealDuration_label with value Spirit Power Steal Duration already exists in ukrainian localization data with value часу духовного ослаблення`
Clearly one of these hasn't been translated yet and the other has, so the implementation was designed to make it use the translated value instead of the untranslated value, instead of whichever one was detected last.

Surxe:
@hassayag Fixed the issue that caused the GHA workflow to fail, should be good to merge now.

Surxe:
Can't find your message regarding --verbose not working, I read it once somewhere then lost it. I think I catch your drift though, since we would just call it as '--verbose' rather than '--verbose true', though I'm still not sure how to accomplish this. The only parameters similar to this are in a different group action and hence have different parameter setups entirely. Could you help me out?

hassayag:
> Can't find your message regarding --verbose not working, I read it once somewhere then lost it. I think I catch your drift though, since we would just call it as '--verbose' rather than '--verbose true', though I'm still not sure how to accomplish this. The only parameters similar to this are in a different group action and hence have different parameter setups entirely. Could you help me out?

And what is it that is different for those parameters? Can you see any different arguments passed in?

src/parser/parser.py
hassayag: Ah it isn't uploaded you're right. I'm not sure the best way to handle that tbh, as even with a successful comparison, we need something more than a log message buried in the gha. But uploading to the wiki is one option no? If there is a change to `meaningful-data.json`, you would see that in your watch list

Surxe:
> And what is it that is different for those parameters? Can you see any different arguments passed in?
Within parameters.py I see that bot_push and the rest in that group have `action='store_true'`, and no `default` parameter. The Think I figured it out. I did not find the `default` parameter's obstruction communicated so I added documentation to the top of `parameters.py` which should help future editors.

hassayag:
> > And what is it that is different for those parameters? Can you see any different arguments passed in?
> > Within parameters.py I see that bot_push and the rest in that group have `action='store_true'`, and no `default` parameter. The Think I figured it out. I did not find the `default` parameter's obstruction communicated so I added documentation to the top of `parameters.py` which should help future editors.

You've spotted the correct params, but its purely just to do with the `action='store_true'`, not the default value. So you just need to set that `action` argument to fix this

Surxe:
> > > And what is it that is different for those parameters? Can you see any different arguments passed in?
> > > Within parameters.py I see that bot_push and the rest in that group have `action='store_true'`, and no `default` parameter. The Think I figured it out. I did not find the `default` parameter's obstruction communicated so I added documentation to the top of `parameters.py` which should help future editors.
> 
> You've spotted the correct params, but its purely just to do with the `action='store_true'`, not the default value. So you just need to set that `action` argument to fix this

Ahh I see. Where can we see docs on what action=store_true means? For now I added this at the top of parameters.py. Thanks for the guidance

hassayag:
> Ahh I see. Where can we see docs on what action=store_true means? For now I added this at the top of parameters.py.

https://docs.python.org/3/library/argparse.html

hassayag (approved):
Working great! For that meaningful stats thing, go ahead and create a git issue so we don't forget about it

PR #94, opened 2024-10-31
Feature/parse generic data

Surxe:
**Overview**
"Parses" `generic_data.vdata`

This parsing process is a new approach such that Valve's structure is not modified in any way. If a structure is detected as different to what's currently in `output-data`, it throws a warning that points to request changes be made on the frontend module `[[Module:GenericData]]` (page not yet created).

The only real parsing that occurs is the removal of prefixes from the keys, like `m_mapMyCoolKey` --> `MyCoolKey`, `flMyCoolKey` -> `MyCoolKey`, and `EMeleeDamage` -> `MeleeDamage`.

This approach shifts the focus from pre-processing (python backend) to post-processing (lua scripting and templates for wiki purposes, external repositories for other purposes), and as a result the repository and its data would be more:
* Accessible to contribution from the perspective of wiki editors
* Easily maintainable by us, allowing for a possibility of our own API, suiting the needs of those seeking a Valve-backed API
* Stable/Robust to change, detecting structural changes more precisely when Valve inevitably modifies them

**Example**
Only 3 data sets within GenericData I have immediate ideas for usage on the wiki, 1 of these 3 is as follows:
```
"ObjectiveParams": {
        "GoldPerOrb": 40,
        "Tier1GoldKill": 175,
        "Tier1GoldOrbs": 100,
        "Tier2GoldKill": 500,
        "Tier2GoldOrbs": 250,
        "BaseGuardiansGoldKill": 85,
        "BaseGuardiansGoldOrbs": 40,
        "ShrinesGoldKill": 0,
        "ShrinesGoldOrbs": 0,
        "PatronPhase1GoldKill": 0,
        "PatronPhase1GoldOrbs": 0
    },
```

After being saved to `output-data`, if I manually change the key `GoldPerOrb` to `SoulPerOrb` (for illustration purposes), save it, then reparse, the warning displayed is
```
*WARNING* A structure within Generic data: {'ObjectiveParams': {'SoulPerOrb': 'The key for this value is missing in one of the data sets'}} is now different. Please verify the changes, and update the frontend page [[Module:GenericData]] accordingly.
```

If simply a value is changed (but is still the same datatype), no warning is raised. 

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/18)_


Surxe:
Uploaded to [[Data:GenericData.json]], module code written on [[Module:GenericData]] and [[Module:ItemData/nav]].

The 3 current uses are:
* ItemPricePerTier, for creating the soul groups in both item nav boxes.
image

* RejuvParams, for various calls on [[Rejuvenator]].
image

* ObjectiveParams, for listing souls gathered from objectives on [[Souls]].
image

Surxe:
Closing and reopening for data repo

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+            invalid_keys = validate_structures(
+                existing_generics, parsed_generics, structure_keys_to_validate
+            )
+            if len(invalid_keys) > 0:
+                print(
+                    '*WARNING* A structure within Generic data:'
+                    + f' {invalid_keys} is now different. Please verify the changes,'
+                    + ' and update the frontend page [[Module:GenericData]] accordingly.'
+                )
+
```
hassayag, line 45: I removed half of `generic-data.json` and it didn't throw a single warning.

Also with how the release pipeline will work automatically, I doubt these messages will even be seen

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+            for i, elem in enumerate(value):
+                if isinstance(elem, dict):
+                    value[i] = remove_prefixes(elem, possible_prefixes)
+
+        new[new_key] = value
+
+    return new
+
+
+def remove_prefix(key, prefix):
```
hassayag, line 121: This belongs in `string_utils.py`, with replacement of the word "key" with "string"

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+                for i, elem in enumerate(datas1[key]):
+                    if isinstance(elem, dict):
+                        more_invalid_keys = validate_structures(elem, datas2[key][i], elem.keys())
+                        if len(more_invalid_keys) > 0:
+                            invalid_keys[key] = more_invalid_keys
+
+    return invalid_keys
+
+
+def remove_prefixes(generic_data, possible_prefixes):
```
hassayag, line 95: This is probably better suited inside of the class, unless it has any use outside of it?

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+                print(
+                    '*WARNING* A structure within Generic data:'
+                    + f' {invalid_keys} is now different. Please verify the changes,'
+                    + ' and update the frontend page [[Module:GenericData]] accordingly.'
+                )
+
+        return parsed_generics
+
+
+def validate_structures(datas1, datas2, structure_keys_to_validate):
```
hassayag, line 49: this could be a json_util as it may be useful elsewhere in other places

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+        # Parse generic data
+        possible_prefixes = ['m_str', 'm_map', 'm_n', 'm_fl', 'm_', 'fl', 'E', 'n']
+        parsed_generics = remove_prefixes(self.generic_data, possible_prefixes)
+
+        # Read existing generic data
+        existing_generics = self._read()
+
+        # If there ie existing data, validate the structure
+        if existing_generics is not None:
+            structure_keys_to_validate = ['ObjectiveParams', 'RejuvParams', 'ItemPricePerTier']
```
hassayag, line 35: constants should go inside a class attr at the top

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+    def _read(self):
+        if not os.path.exists(self.OUTPUT_DIR):
+            return None
+
+        # Read existing generic data from file
+        return json_utils.read(self.OUTPUT_DIR)
+
+    def run(self):
+        # Parse generic data
+        possible_prefixes = ['m_str', 'm_map', 'm_n', 'm_fl', 'm_', 'fl', 'E', 'n']
```
hassayag, line 27: same thing as `structure_keys_to_validate`

src/parser/parsers/generics.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    Lightly parse the generic data to have more readable keys.
+
+    Validate structures used by the frontend to identify
+    possible failure points in the case they are modified
+    by Valve.
+    """
+
+    def __init__(self, output_dir, generic_data):
+        self.OUTPUT_DIR = output_dir
```
hassayag, line 15: name this generic_data_dir as output_dir at the moment is widely used to refer to the environment var

hassayag (changes requested):
Nicely automated with all the prefix checks 👌

src/parser/parsers/generics.py
Surxe: Good catch, should be fixed now; before it was only checking if all the keys in 1 dict existed in the other. Now it should be checking in both directions. Didn't think to test that part earlier for some reason, was only testing changing the name of keys instead of outright removing them, and it not checking in both directions meant I was accidentally picking the flawless direction each time lol.

Also recall that for the 1st-layer of the dict, its only looking at keys in `structure_keys_to_validate`

Didn't think about the release pipeline. We need to ensure it handles errors well, as we should expect them to occur often when Valve changes structures. What idea do you propose for this? The first idea that pops into my head is a `output-data/warnings.txt/` file. `deadbot.py` would remove this file on every run to ensure it only exists if there's warnings from the previous runtime.

src/parser/parsers/generics.py
Surxe: good point

hassayag (approved):
Looks good. Remember to update with develop first, as `output-data` shouldn't be tracked

hassayag:
> > Looks good. Remember to update with develop first, as `output-data` shouldn't be tracked
> 
> I meant to ask about this actually - for some reason output-data is in the file diff, but there's no diff in .gitignore, so shouldn't there be no diffs in output-data at all? Any idea?

.gitignore just stops you committing those files locally, so I think this would have come from bad merge conflicts

Surxe:
> > > Looks good. Remember to update with develop first, as `output-data` shouldn't be tracked
> > 
> > 
> > I meant to ask about this actually - for some reason output-data is in the file diff, but there's no diff in .gitignore, so shouldn't there be no diffs in output-data at all? Any idea?
> 
> .gitignore just stops you committing those files locally, so I think this would have come from a bad merge

I'm still confused- what do you propose the solution is?

hassayag:
> > > > Looks good. Remember to update with develop first, as `output-data` shouldn't be tracked
> > > 
> > > 
> > > I meant to ask about this actually - for some reason output-data is in the file diff, but there's no diff in .gitignore, so shouldn't there be no diffs in output-data at all? Any idea?
> > 
> > 
> > .gitignore just stops you committing those files locally, so I think this would have come from a bad merge
> 
> I'm still confused- what do you propose the solution is?

PR #97, opened 2024-11-05
Bugfix/fix stake t3 description

hassayag:
_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/14)_


Surxe:
@hassayag I get the following error when parsing
image

If I make an exception for it to skip; I see theres 1 more similar case too, though not a 3rd one.

hassayag:
> @hassayag I get the following error when parsing
image
If I make an exception for it to skip; I see theres 1 more similar case too, though not a 3rd one.

Are you on latest game files? Working fine for me

Surxe (approved):
Looks good!

PR #100, opened 2024-11-07
Feature/remove hero bound abilities

Surxe:
Removes duplicate data from BoundAbilities in hero data that is sufficiently in ability data instead. The only remaining keys are Key and Name(english).

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/15)_


src/parser/parser.py
```diff
@@ -141,8 +142,20 @@ def _parse_heroes(self, parsed_abilities):
             json_utils.sort_dict(parsed_meaningful_stats),
         )
 
+        stripped_heroes = dict()
+        # Remove irrelevant data from BoundAbilities in HeroData
+        for hero_key, hero_value in copy.deepcopy(parsed_heroes).items():
```
hassayag, line 147: looks like hero_value is not being reassigned anywhere, so is a copy strictly necessary?

Surxe: I tried without first; if you set a dict value within stripped_heroes to be the dict value from parsed_heroes, they point to the same address, causing the nested values within BoundAbilities to be shared. Couldn't get it to work without the deepcopy. Lmk if you can think of a cleaner solution though

hassayag: Oh yeah it assigns it to stripped heroes, which is then manipulated. Thats fine then

hassayag (approved):
Looks good, just make sure the wiki doesn't rely on this data

PR #102, opened 2024-11-08
Bugfix/remove panel tags and added bolding for other span classes

hassayag:
They've started using `<Panel>` tags and other `<span>` classes to ability descriptions. "Panel" is meant to represent an icon, but for now it has been removed. Other span classes have been given some simple bold/italic styling, as colouring them might be over the top.

Since it was causing `<Panel ...></Panel>` tags to show on the new Trapper hero, I've pushed the changes to Lang_en already

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/17)_


Surxe (approved):
Looks simple enough, nice!

PR #127, opened 2024-11-22
Hotfix/boon mapping

Surxe:
Map boons and max charges; and temp fix for IGNORE_KEYS
The temporary change in string_utils will want to be reverted later on when a more permanent solution is found.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/37)_


hassayag (approved):
We can work out the problematic ignore list another time. Thanks for the 🔥fix

PR #111, opened 2024-11-10
Feature/herolab patch note parser

Surxe:
Hero Lab patch notes are now parsed through their location in the gamefiles, rather than added to input-data manually.

* Removed the comments and readme for adding to input-data.
* Reformatted dates from forum changelogs from mm-dd-yyyy to yyyy_mm_dd, which are now consistent with gamefile date formats. This allows them to be sorted by date accurately in `changelog_configs.json` (formerly named `changelogs.json`).
* Fixed the TypedDicts accidentally describing the final format that `parse_changelogs` uses with Tagging, as opposed to it's actual format.

Still feel's a little scrappy, so lmk if you see any immediate areas of improvement. I'll need this merged in before I work on integrating the manifest-id versioning branch with the recent changelog updates.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/25)_

PR #130, opened 2024-11-25
Feature/herolab patch note

Surxe:
Ported from original PR #111 

Hero Lab patch notes are now parsed through their location in the gamefiles, rather than added to `input-data` manually. 

* `changelog_configs.json` is now saved in `input-data` instead of `output-data` temporarily. This will be developed more in a future PR.
* Removed the comments and readme for adding manually to `input-data`.
* Reformatted dates from forum changelogs from mm-dd-yyyy to yyyy_mm_dd, which are now consistent with gamefile date formats. This allows them to be sorted by date accurately in `changelog_configs.json` (formerly named `changelogs.json`).
* Fixed the TypedDicts accidentally describing the final format that `parse_changelogs` uses with Tagging, as opposed to it's actual format.
* Corrected `Powerup` tag to `PowerUp`

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/40)_


hassayag (approved):
Beautiful 🚀

PR #137, opened 2024-11-30
fix: fixed scaling data being missed off output data for some attributes

hassayag:
_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/44)_


Surxe (approved):
Looks simple enough, nice!

PR #138, opened 2024-11-30
fix: save rss changelogs to input data

hassayag:
The new fetching process loads `input-data/changelogs` files first. Any output data is then also written to input data, as `input-data` is to be treated as a persistent storage - this is handled by the deploy script, in which it commits any `input-data` changes. 

Since the RSS feed can only see recent changelog data, the rest will now be made up by this `input-data`. Therefore `output-data` can be consistently reproduced from nothing. Because of that consistency, deadlock-data has been reverted back to the simple `data` directory, which reflects what `output-data` would look like locally

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/45)_


Surxe (approved):
I still have roughly 2 hours of available effort a day due to sickness, so I skimmed very lightly, but this looks very good! Like the idea, even if it means input-data may have slightly more diffs, I think its well worth. Also I like the simplified external calls for changelog fetcher
image

PR #139, opened 2024-12-04
Feature/mcginnis rev firerate calc

Surxe:
McGinnis' weapon "Revs" or "Spins" up, decreasing bullet spread and more importantly, increasing her fire rate. This PR adds proper calculation for the 3 respective attributes, "SpinAcceleration", "SpinDeceleration", and "RoundsPerSecondAtMaxSpin".

Only "RoundsPerSecondAtMaxSpin" is used in dps calculation because the DPS calc assumes already being at max spin.

Uploaded to [[Data:HeroData.json]] and confirmed it works on [[Hero Comparison]] page as expected.

See the commit message regarding deadlock-data in .gitignore

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/46)_

### DIFF
--- a/src/parser/parsers/heroes.py
+++ b/src/parser/parsers/heroes.py
@@ -201,11 +201,13 @@ class HeroParser:
         }
 
         if 'm_bSpinsUp' in w and w['m_bSpinsUp'] == 1:
-            weapon_stats.update({
-                'RoundsPerSecondAtMaxSpin': 1 / w['m_flMaxSpinCycleTime'],
-                'SpinAcceleration': w['m_flSpinIncreaseRate'],
-                'SpinDeceleration': w['m_flSpinDecayRate'],
-            })
+            weapon_stats.update(
+                {
+                    'RoundsPerSecondAtMaxSpin': 1 / w['m_flMaxSpinCycleTime'],
+                    'SpinAcceleration': w['m_flSpinIncreaseRate'],
+                    'SpinDeceleration': w['m_flSpinDecayRate'],
+                }
+            )
 
         dps_stats = self._get_dps_stats(weapon_stats)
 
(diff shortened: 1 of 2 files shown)


src/parser/parsers/heroes.py
```diff
@@ -227,7 +236,9 @@ def _get_dps_stats(self, weapon_stats):
             'ReloadDelay': weapon_stats['ReloadDelay'],
             'ReloadTime': weapon_stats['ReloadTime'],
             'ClipSize': weapon_stats['ClipSize'],
-            'RoundsPerSecond': weapon_stats['RoundsPerSecond'],
+            'RoundsPerSecond': weapon_stats['RoundsPerSecondAtMaxSpin']
```
hassayag, line 239: This has skewed the dps way higher than it actually is, as it implies mcginnis fires at max spin all the time. Instead the spin stats should be separate to this and just considered in the DPS calculation

hassayag (changes requested):

src/parser/parsers/heroes.py
hassayag: I just saw this in the description 

> Only "RoundsPerSecondAtMaxSpin" is used in dps calculation because the DPS calc assumes already being at max spin.

Which DPS calc assumes it, the one in game?

Surxe: I'm not sure which one it displays in game, it might be dynamic to how revved up you are. 

I've intentionally set it up to use fully revved up RPS because its up to 2.5x RPS and after only 2 seconds of firing, making fully-revved up dps much more accurate as a reference to her average dps than non-revved up. 

> Instead the spin stats should be separate to this and just considered in the DPS calculation
The spin stats already only factor into DPS. If users want to know her un-revved dps, they can quickly determine this as well.

hassayag (approved):

PR #141, opened 2024-12-07
Feature/changelog ids

Surxe:
When the changelog pipeline was reworked in #84, we switched over to using the forum id as the changelog id that its stored in. After hero lab's released (which don't have a forum id), we needed to start making up changelog id's. Now, the changelog id is made up but in inconsistent formats with no intended usage on the frontend.

This PR aims to use the date as the changelog-id in preparation of using them for wiki page names, i.e. `Data:changelog_id.json` and `Update:<changelog_id>`. However, unlike before, if multiple changelogs are released on the same day, they will have a number appended at the end. For example, most changelogs id's will look like `2024-12-07` or `2024-12-07_HeroLab`, however, on October 18th they released two patches, so the id's would be `2024-10-18` and `2024-10-18-1`.

Additionally, each config in `changelog_configs` now has boolean flag `is_hero_lab`. Though it isn't used anywhere yet, I expect this to make it easier to display herolab or non-herolab specific patches on the wiki's side.

TODOS:
- ~~Add isHeroLab to changelog configs~~
- ~~Cleanup code a bit in `_create_changelog_id`. I'm not very happy with the elegance of how the solution is written, buut it works lol.~~ ~~ok its a bit better now~~ ok its a lot better now
- ~~Test it more thoroughly when next patch comes out~~ tested scenario where two patches are released on same day, and scenario where two patches are released on different days, both passed
- ~~Swap underscores for hyphens in changelog-ids~~ this was done because underscores are replaced with a space in wiki page names, otherwise making the page names appear awkwardly as `2024 12 10`

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/47)_


Surxe:
@hassayag Ready for review- be sure to reread the top conversation post as I made many edits. Also, not sure which label to put for this one, infra or just enhancement? It minorly changes infrastructure of stored changelog data, so somewhere in between.

src/changelogs/fetch_changelogs.py
```diff
@@ -286,10 +293,40 @@ def _process_local_changelogs(self, changelog_path):
...
+        # Else same date already exists
+
+        # If the forum id is the same, use the same changelog id 
+        # which will update the existing record
+        if existing_config['forum_id'] == forum_id:
+            return id
+        # Else forum id's are different, so different patches on the same day
+
+        # Recursively check if the next id is available
+        return self._create_changelog_id(date, forum_id, i + 1)
```
hassayag, line 324: Neat solution!

hassayag (approved):
Great stuff. Also satisfying to have the changelogs in chronological order now 🙌

PR #142, opened 2024-12-09
fix: skip heroes without upgrade data

hassayag:
_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/48)_


hassayag:
closed due to issue with branch head

PR #154, opened 2024-12-22
Bugfix/misc changelog fixes

Surxe:
* Some tags in `tags.py` were reordered just to change some examples from `{{PageRef|Crate|alt_name=crate}}s` to `{{PageRef|Crate|alt_name=crates}}`. While it doesn't really change anything, it does help readability, and improves consistency with ordering in the several lists in `tags.py`. 
* Empty changelog lines no longer skipped. This allows the spacing to be setup properly immediately on changelog pages, without humans needing to revise the spacing.
* Latest changelog from 12-21-2024 included in input-data.
* Now avoids hyperlinking to [[Other]] just because its the default tag
* Removed the "HeroLab <heroname>" tags, its now just "<heroname>" tag. The changelogs are able to be sectioned off by hero lab and non-herolabs now so these extra tags are no longer needed. 
* Each config in `changelog_configs` now has a `sequence_id` that starts at 0 for each day, designating multiple patches on the same day. This is also used for adding `frontfacing_date_id` in the format `May 3, 2024` and `May 3, 2024-1` which will be the patch names we use on the wiki.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/54)_


Surxe:
Converted it to a draft earlier while I added some other minor corrections, its back to ready for review now

hassayag (changes requested):
Left comments on the deadlock data PR as well

PR #156, opened 2024-12-23
Feature/auto increment pyprojecttoml version

hassayag:
Created auto-increment of pyproject.toml version to ensure unique versioning when pushing to wiki from `develop`.
Actual version increments (not just betas) will come in a future PR once the release process is more automated

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/55)_


Surxe (approved):
Good idea, I was wondering about this myself just recently haha

PR #159, opened 2024-12-27
Use new discord link

Surxe:
A standalone server was created for the wiki, we can now use this link in the readme instead of the deadlock community discord.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/57)_


hassayag (approved):
Good shout 👌

Issue #63, opened 2024-10-03
Standardise "Disabled" flag on items and heroes


hassayag:
There is use of `Disabled` and `IsDisabled` across hero and item data. Let's stick with just `Disabled` to keep it consistent

kiara-abbey:
Hi, I'm a college student in an open-source software course. I can work on this!!! Would i just be editing the json files?

hassayag:
Hey! All json files in `output-data/json` are generated from the source code and are not editted manually.

We currently have two disabled flags, one for items, one for heroes. If you have a look around their parsers in [here](https://github.com/deadlock-wiki/deadbot/tree/master/src/parser/parsers), you will see they each use a different name for their disabled flag. 

When you have completed that, you will need to run the parser locally, which should then show changes in some of the `output-data` files.

Just be sure to go through the README first to get setup if you haven't already. Hope that helps!

kiara-abbey:
This super helps! :) Thanks so much.

B13rg:
Related pr: #49

hassayag:
@kiara-abbey is this still with you?

PR #163, opened 2024-12-29
Bugfix/missing is disabled keys

Surxe:
- Addresses #63 
- Parses m_bDisabled to IsDisabled for Ability Data (independent to Ability Card parsing)

This standardization will then be used by the new deadlock-wiki/pages repository.

I tested it locally and got IsDisabled key seen on ability-data.json output data

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/62)_

### DIFF
--- a/src/parser/parsers/abilities.py
+++ b/src/parser/parsers/abilities.py
@@ -26,7 +26,7 @@ class AbilityParser:
             ability_data = {
                 'Key': ability_key,
                 'Name': self.localizations.get(ability_key, None),
-                'IsDisabled': ability.get('m_bDisabled', None),
+                'IsDisabled': ability.get('m_bDisabled', False),
             }
 
             stats = ability['m_mapAbilityProperties']
--- a/src/parser/parsers/abilities.py
+++ b/src/parser/parsers/abilities.py
@@ -26,6 +26,7 @@ class AbilityParser:
             ability_data = {
                 'Key': ability_key,
                 'Name': self.localizations.get(ability_key, None),
+                'IsDisabled': ability.get('m_bIsDisabled', False),
             }
 
             stats = ability['m_mapAbilityProperties']
--- a/src/parser/parsers/abilities.py
+++ b/src/parser/parsers/abilities.py
@@ -26,7 +26,7 @@ class AbilityParser:
             ability_data = {
                 'Key': ability_key,
                 'Name': self.localizations.get(ability_key, None),
-                'IsDisabled': ability.get('m_bIsDisabled', True),
+                'IsDisabled': ability.get('m_bDisabled', None),
             }
 
             stats = ability['m_mapAbilityProperties']
--- a/src/parser/parsers/abilities.py
+++ b/src/parser/parsers/abilities.py
@@ -26,7 +26,7 @@ class AbilityParser:
             ability_data = {
                 'Key': ability_key,
                 'Name': self.localizations.get(ability_key, None),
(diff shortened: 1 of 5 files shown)


src/parser/parsers/items.py
```diff
@@ -104,7 +104,7 @@ def _parse_item(self, key):
         return parsed_item_data
 
     def _is_disabled(self, item):
-        is_disabled = False
+        is_disabled = True
```
hassayag, line 107: Why is this now defaulted as `True`?

src/parser/parsers/abilities.py
```diff
@@ -26,6 +26,7 @@ def run(self):
             ability_data = {
                 'Key': ability_key,
                 'Name': self.localizations.get(ability_key, None),
+                'IsDisabled': ability.get('m_bIsDisabled', True),
```
hassayag, line 29: similarly, why is this default to `True`?

hassayag (changes requested):

src/parser/parsers/items.py
hassayag: I only see like 15 items with `"IsDisabled": false`, so I think defaulting to False is correct here

src/parser/parsers/abilities.py
hassayag: Just noticed every ability is disabled because this should be `m_bDisabled`. I still don't think this should default to True though

Surxe: Oh, that explains my failed state of reference then

Surxe: Decompiled abilities with Disabled=:
* true, 59 - looks correct, all of these are not in game
* false, 19 - several in here are definitely enabled, but some i'm pretty certain are not in game
In parsed data when I set default=none, ~180 are null, including ones I definitely know are in game like infernus' afterburn

Items:
* true, 72 - never heard of any of these so looks correctly disabled
* false, 17 - they all look correctly enabled, but there should be more than 17
* null, 117 - some of these are in game like refresher and some are not

Idek anymore, can't find any pattern. I pushed the commit to default both to None if you want to take a look.

hassayag: I'd be interested in any examples you found where the flag wasn't found, but you say are not in game

Surxe: > I'd be interested in any examples you found where the flag wasn't found, but you say are not in game

All of bomber's abilities, i.e. `ability_bomber_ability03`, bomber hero's flags are:
```
        "InDevelopment": true,
        "InHeroLabs": false,
        "IsDisabled": true,
        "IsRecommended": false,
        "IsSelectable": true,
```
So I suppose the definition of what's "in game" is a bit blurry, at least for hero abilities.
Also notably `ability_death_tax`, which isn't attached to any hero in decompiled data so I presume its an item.

hassayag: I guess it depends on how they use their disabled flags. From your evidence, the majority of items/abilities that don't have a disabled flag seem to be in the game, so i think defaulting to False is correct here

Surxe: Well, it seems that the problem is now that regardless of how we default it, there will be items that are enabled that aren't attainable in game. Is there another local file perhaps that determines if they are attainable, since clearly its not using the IsDisabled key? Your right though that defaulting to False is definitely the lesser evil- I'll set the default back to False and we can merge it, then hopefully find something more concrete for longevity sake.

hassayag (approved):

PR #182, opened 2025-08-10
feat: add Spirit/Boon scaling to items, fix Mermaid output 

LVL1024:
### What changed

- **Updated `items.py`** to extract Spirit and Boon scaling **only when `m_flStatScale` is present**, preventing false positives.
- **Added configurable `scaling_type_map`** for future scaling types.
- **Improved type safety, logging, and edge-case handling.**
- **Fixed return type**: `items.ItemParser.run()` now returns a `MermaidDiagram` object instead of a raw string.

### Compatibility fix in `parser.py`
- Changed `f.write(item_component_chart)` → `f.write(str(mermaid_diagram))` to match the new return type.
- Ensures `item-component-tree.txt` remains unchanged and compatible with Mermaid.js rendering.

### DIFF
--- a/src/parser/parser.py
+++ b/src/parser/parser.py
@@ -225,18 +225,19 @@ class Parser:
 
     def _parse_items(self):
         logger.trace('Parsing Items...')
-        (parsed_items, item_component_chart) = items.ItemParser(
+        (parsed_items, mermaid_diagram) = items.ItemParser(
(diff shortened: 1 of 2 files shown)


hassayag:
Merged into new branch on https://github.com/deadlock-wiki/deadbot/pull/183 to resolve CI issues

PR #183, opened 2025-08-11
feat: add spirit scaling for item data

hassayag:
### What changed

- **Updated `items.py`** to extract Spirit and Boon scaling **only when `m_flStatScale` is present**, preventing false positives.
- **Added configurable `scaling_type_map`** for future scaling types.
- **Improved type safety, logging, and edge-case handling.**
- **Fixed return type**: `items.ItemParser.run()` now returns a `MermaidDiagram` object instead of a raw string.

### Compatibility fix in `parser.py`
- Changed `f.write(item_component_chart)` → `f.write(str(mermaid_diagram))` to match the new return type.
- Ensures `item-component-tree.txt` remains unchanged and compatible with Mermaid.js rendering.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/72)_

### DIFF
--- a/src/parser/parsers/items.py
+++ b/src/parser/parsers/items.py
@@ -2,79 +2,45 @@ from python_mermaid.diagram import MermaidDiagram, Node, Link
 import utils.string_utils as string_utils
 import parser.maps as maps
 from loguru import logger
-import math
-from typing import Dict, Any, Optional, Tuple, List
-
 
 class ItemParser:
-    """Parses item data from game files, including base stats and scaling (Spirit/Boon)."""
-
-    DEFAULT_SCALING_TYPE_MAP = {
-        "ETechPower": "Spirit",  # Spirit scaling
-        "ELevelUpBoons": "Boon",  # Boon scaling
-        # Add new mappings here as needed; can be overridden via constructor
+    nodes = []
+    links = []
+    
+    # Define scaling types that we want to extract
+    # This makes it easy to add new scaling types in the future
+    SCALING_TYPES = {
+        'ETechPower': 'Spirit Scaling',
+        'ELevelUpBoons': 'Boon Scaling'
     }
-
-    def __init__(
-        self,
-        abilities_data: Dict[str, Any],
-        generic_data: Dict[str, Any],
-        localizations: Dict[str, str],
-        scaling_type_map: Optional[Dict[str, str]] = None,
-    ):
-        """
-        Initialize the parser.
-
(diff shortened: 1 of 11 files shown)


LVL1024:
Ok, I think I got it working @hassayag

src/parser/parsers/items.py
```diff
@@ -72,15 +69,26 @@ def _parse_item(self, key):
...
+                # Keep the original string value for backward compatibility
+                parsed_item_data[attr_key] = attr['m_strValue']
+                
+                # Check for scaling information
+                scaling_info = self._extract_scaling_info(attr, key, attr_key)
+                if scaling_info:
+                    scaling_type, scaling_value = scaling_info
+                    # Create a new key with _scaling suffix
+                    scaled_key = f"{attr_key}_scaling"
+                    parsed_item_data[scaled_key] = {scaling_type: scaling_value}
```
hassayag, line 91: I'd suggest something like this for the structure of an attribute which has scaling. Similar to how it is for ability data
```json 
{
    "Heal": {
        "Value": 35,
        "Scale": {
            "Value": 1.0,
            "Type": "spirit"
        }
    }
}
```

And you'll see an existing map for the scaling types here if that works for you https://github.com/deadlock-wiki/deadbot/blob/393a676b4b5b991665a985393e830923f1df609c/src/parser/maps.py#L173-L196

hassayag (commented):
Great work! I've suggested something else for the scaling data structure, let me know what you think

src/parser/parsers/items.py
LVL1024: I like that, we should go with that.

src/parser/parsers/items.py
```diff
@@ -69,43 +62,84 @@ def _parse_item(self, key):
...
+
+        try:
+            value = float(raw_value)
+            if math.isnan(value) or math.isinf(value):
+                return None
+        except (ValueError, TypeError):
+            return None
+
+        return {
+            "Value": value,
```
hassayag, line 136: Have a look at the output in https://github.com/deadlock-wiki/deadlock-data/pull/72/files#diff-ee45ad4cde47a873f0448c021e4dba52cc6e54b13b7c84770d25f43feeeee0fcR1148-R1151

this base value is currently set to the same as the scale

LVL1024: I believe I fixed it now.

PR #203, opened 2025-09-01
feat: Add Hero Type parsing

raimannma:
This pull request introduces support for parsing and including hero type information in the hero data extracted by the parser.


src/parser/parsers/heroes.py
```diff
@@ -144,6 +145,18 @@ def _get_meaningful_stats(self, all_hero_stats):
 
         return meaningful_stats
 
+    def _parse_hero_type(self, hero_value):
```
hassayag, line 148: No clue why, but if you see the [deadlock-data side](https://github.com/deadlock-wiki/deadlock-data/pull/82/files#diff-7c1ef3c763f57b5ec05ba132508c5ad0390a4c3c69cfa5754a1ad2fdb03f1f90R2686), it's showing "Braw" instead of "Brawler"

hassayag (changes requested):
Solid first PR, one issue I can see on the parser output

Issue #72, opened 2024-10-11
Correct endlines in localization data


Surxe:
Currently, endlines in decompiled localization data `\n` are parsed into `\"n`. Instead, replace them with `<br>` in `parsers/localization` such that the wikitext reads it correctly.

See deadlocked.wiki/User:Sur/Abrams#Introduction as an example of how it appears currently.

PR #218, opened 2025-09-10
fix: Improve item/localization parsing, hero stat formatting, and update dependencies

LVL1024:
- **Localization**: Replaced the localization parser to correctly convert newlines (`\n`) to `<br>` and prevent the corruption of HTML tags and other data.
- **Hero Stats**: Formats `CritDamageBonusScale` as a percentage (e.g., 1.0 → 0%) to match in-game display and rounds stat calculations to prevent float precision errors.
- **Item Data**: Converts attribute values to numbers, filters out zero-value attributes, and adds safety checks to prevent overwriting core fields.
- **Changelogs**: Fixes a bug that could cause data loss when parsing changelogs.
- **Dependencies**: Updates pandas to v2.2.2 and replaces the deprecated `applymap()` with `map()`.

---
_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/95)_

### DIFF
--- a/src/parser/parsers/heroes.py
+++ b/src/parser/parsers/heroes.py
@@ -34,8 +34,14 @@ class HeroParser:
 
                 # Change formatting on some numbers to match whats shown in game
                 hero_stats['StaminaCooldown'] = 1 / hero_stats['StaminaRegenPerSecond']
-                hero_stats['CritDamageReceivedScale'] = (hero_stats['CritDamageReceivedScale'] - 1) * 100
-                hero_stats['CritDamageBonusScale'] = round((hero_stats['CritDamageBonusScale'] - 1) * 100, 2)
+                
+                # Convert scale values to percentages and rename keys for clarity
+                received_scale = hero_stats.pop('CritDamageReceivedScale')
+                hero_stats['CritDamageReceivedPercent'] = round((received_scale - 1) * 100, 2)
+                
+                bonus_scale = hero_stats.pop('CritDamageBonusScale')
+                hero_stats['CritDamageBonusPercent'] = round((bonus_scale - 1) * 100, 2)
+                
                 hero_stats['TechRange'] = hero_stats['TechRange'] - 1
                 hero_stats['TechDuration'] = hero_stats['TechDuration'] - 1
                 hero_stats['ReloadSpeed'] = hero_stats['ReloadSpeed'] - 1


LVL1024:
Tried to fix #72 , but the diffs still showed those `\\'` slashes because another part of the process re-escapes the text after localization, and I couldn't figure out how to fix it.

src/decompiler/localization.py
```diff
@@ -15,14 +28,39 @@ def process_files(input_folder, output_folder):
...
+                    right = unescaped_right.replace('\n', '<br>').strip()
 
                     out[left] = right
+                else:
+                    # Log lines that look like they should have matched, which could indicate a format change.
+                    if '"' in stripped_line:
+                        logger.debug(
+                            f'Skipping malformed localization line in {filename}: '
+                            f'"{stripped_line}"'
+                        )
```
hassayag, line 61: We can probably not log these malformed lines as there are a lot. It looks like they are either commented out, or things like "lang", so not the most useful

src/parser/parsers/items.py
```diff
@@ -74,12 +74,21 @@ def _parse_item(self, key):
 
             scaling_data = self._extract_scaling(attr, key, attr_key)
 
+                        # Create a set of keys that should not be overwritten
+            protected_keys = {'Name', 'Description', 'Cost', 'Tier', 'Activation', 'Slot', 'Components', 'TargetTypes', 'ShopFilters', 'IsDisabled'}
+            
+            # If the attribute key is one of the protected keys, skip it to prevent overwriting.
+            if attr_key in protected_keys:
```
hassayag, line 81: Is there a reason we need to manually ignore these keys? I see the same output with this removed

src/parser/parsers/items.py
```diff
@@ -74,12 +74,21 @@ def _parse_item(self, key):
```
hassayag, line 74: while we are here, thoughts on updating the `Cost` and `Tier` attributes to be numbers instead of strings? (see above this line)

main.sh
```diff
@@ -7,7 +7,7 @@ if [ -f ".env" ]; then
 fi
 
 # Configured completly from env vars:
-python3 src/deadbot.py
+poetry run deadbot
```
hassayag, line 10: After you changed the pandas version, I noticed issues with my package versions because we were not using the poetry venv to run it, so this should hopefully fix that from now on

hassayag (changes requested):
Great work 👍 few suggestions and questions

src/parser/parsers/items.py
LVL1024: You're right. I've updated it to store `Cost` and `Tier` as numeric types.

src/decompiler/localization.py
LVL1024: Done. Removed the logging.

src/parser/parsers/items.py
```diff
@@ -74,12 +74,21 @@ def _parse_item(self, key):
...
+
             if scaling_data:
-                # Place the structured attribute directly on the item object
                 parsed_item_data[attr_key] = scaling_data
             elif 'm_strValue' in attr:
-                # Otherwise, it's a simple key-value pair at the top level
-                parsed_item_data[attr_key] = attr['m_strValue']
+                value = num_utils.assert_number(attr['m_strValue'])
+                # Only filter if the value is a number and it is zero
+                if isinstance(value, (int, float)) and value == 0:
```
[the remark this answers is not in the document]
LVL1024: I believe in Python the expression `0 == 0.0` evaluates to `True`. So the `value == 0` check should catch both. I ran it and checked the diffs just to be sure, and it looks like both are being filtered out as intended.

src/parser/parsers/items.py
LVL1024: You're right, that check isn't needed anymore. It was a safeguard I added a while back when the parser was accidentally nulling out some of those main attributes. I removed it.

src/parser/parsers/heroes.py
```diff
@@ -35,6 +35,7 @@ def run(self):
                 # Change formatting on some numbers to match whats shown in game
                 hero_stats['StaminaCooldown'] = 1 / hero_stats['StaminaRegenPerSecond']
                 hero_stats['CritDamageReceivedScale'] = (hero_stats['CritDamageReceivedScale'] - 1) * 100
+                hero_stats['CritDamageBonusScale'] = round((hero_stats['CritDamageBonusScale'] - 1) * 100, 2)
```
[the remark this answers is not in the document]
hassayag: Yep exactly that

PR #244, opened 2025-10-21
fix: skip wiki authentication during dry run

LVL1024:
Fixes `--dry-run` so it actually works locally without needing the bot's credentials.

Previously, it tried to log in no matter what, which would error out if you didn't have the secrets set up on your machine. Now it just skips the auth step entirely during a dry run.

Makes local testing way easier and doesn't affect CI.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/121)_


hassayag (approved):
Good idea, one fewer barrier for fully testing uploads 👌

PR #252, opened 2025-11-03
feat: add parsing for alternate fire weapon data

LVL1024:
This PR overhauls the hero weapon parser to add support for alternate fire stats and improve the data structure.

-   **Adds support for alternate fire stats**, storing them under `Weapon.AltFire`.
-   **Breaking Change:** Nests all weapon stats under a `Weapon` key for better organization. Will need to update Wiki's modules that use it.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/128) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/parsers/heroes.py
+++ b/src/parser/parsers/heroes.py
@@ -190,47 +190,47 @@ class HeroParser:
         stats = {}
 
         # Safely parse all weapon attributes using .get() to prevent KeyErrors
-        raw_bullet_speed = w.get('m_flBulletSpeed')
+        raw_bullet_speed = weapon_info.get('m_flBulletSpeed')
         stats['BulletSpeed'] = raw_bullet_speed / ENGINE_UNITS_PER_METER if raw_bullet_speed is not None else None
 
         # Core Stats
-        stats['BulletDamage'] = w.get('m_flBulletDamage', 0)
-        stats['RoundsPerSecond'] = 1 / w['m_flCycleTime'] if w.get('m_flCycleTime') and w['m_flCycleTime'] > 0 else 0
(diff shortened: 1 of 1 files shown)


src/parser/parsers/heroes.py
```diff
@@ -165,76 +174,124 @@ def _parse_hero_type(self, hero_value):
...
+        if primary_slot in bound_abilities:
+            weapon_prim_id = bound_abilities[primary_slot]
+            if weapon_prim_id in self.abilities_data and 'm_WeaponInfo' in self.abilities_data[weapon_prim_id]:
+                primary_ability_data = self.abilities_data[weapon_prim_id]
+                weapon_stats = self._parse_weapon_stats(primary_ability_data['m_WeaponInfo'])
+
+                # The primary weapon name/description key is constructed from the hero's key, not its own ability ID.
+                # e.g., hero_shiv -> citadel_weapon_hero_shiv_set
+                weapon_stats['WeaponName'] = f"citadel_weapon_hero_{hero_key.replace('hero_', '')}_set"
+                weapon_stats['WeaponDescription'] = weapon_stats['WeaponName'] + '_desc'
```
hassayag: Similar to ability cards, I think we can use "NameKey" and "DescKey" now it is implicitly a part of the weapon and is referencing a localization key, rather than the actual string

hassayag (changes requested):
Looks pretty good, and much easier to read with it under a `Weapon` attribute

PR #263, opened 2025-11-18
refactor: clean up npc parser logic and move maps/utils to shared files

LVL1024:
### Key Changes
- **Extracted Utilities:** Moved `_deep_get` and `_read_value` to `json_utils` as `deep_get` and `read_value` to make them globally available for other parsers.
- **Centralized Configuration:** Moved NPC modifier maps (Intrinsic, Aura, Rebirth) from inline code to `maps.py` to decouple data from logic.
- **Modularized Logic:** Broke down monolithic parsing methods (e.g., `_parse_walker`) into smaller, reusable components (`_parse_walker_resistances`, `_parse_walker_aura`).
- Replaced hardcoded index access for Rejuvenator stats with key-based mapping to prevent breakages from upstream data reordering.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/136) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/input-data/changelogs/raw/2025-11-21.txt
+++ b/input-data/changelogs/raw/2025-11-21.txt
@@ -316,3 +316,21 @@ Cooldown Reduction Changes:
 - Yamato: Power Slash medium charge damage from 60% to 50%
 - Yamato: Crimson Slash cooldown increased from 11.5s to 15s
 - Yamato: Crimson Slash T3 increased from -5.75s Cooldown to -9s
+---
+- Gun base damage reduced by 5% (not growth)
+- Medic Pack missing health % reduced from 10% to 8%
+- Stamina regen returned from 5.25s to 5s
+- Neutrals bounty increased by 5%
+- Warden: Bullet damage growth per boon reduced from 0.53 to 0.47
+- Drifter: Base health reduced from 750 to 675
+---
+- Fixed a bug with falloff calculation that was often causing guns to do much more damage than intended. We'll be monitoring to do subsequent adjustments after this.
+---
+- Hero kill bounty split efficiency for 1/2/3/4/5/6 players changed from 100/100/90/84/65/72% to 100/100/85/70/55/50%
+- Guardians HP increased from 4250 to 5500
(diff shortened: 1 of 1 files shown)


src/parser/parsers/npc_units.py
```diff
@@ -375,13 +374,13 @@ def _parse_midboss(self, data, npc_key=None):
...
-            'GoldReward': self._read_value(data, 'm_flGoldReward'),
-            'GoldRewardBonusPercentPerMinute': self._read_value(data, 'm_flGoldRewardBonusPercentPerMinute'),
-            'DamagedByAbilities': self._deep_get(data, 'm_bDamagedByAbilities'),
-            'DamagedByBullets': self._deep_get(data, 'm_bDamagedByBullets'),
-            'MinigameDuration': self._read_value(data, 'm_flVaultMiniGameTime'),
-            'MinigameHitWindow': self._read_value(data, 'm_flVaultMiniGameHitWindow'),
+            'RetaliateDamage': json_utils.read_value(data, 'm_flRetaliateDamage'),
+            'GoldReward': json_utils.read_value(data, 'm_flGoldReward'),
+            'GoldRewardBonusPercentPerMinute': json_utils.read_value(data, 'm_flGoldRewardBonusPercentPerMinute'),
+            'DamagedByAbilities': json_utils.deep_get(data, 'm_bDamagedByAbilities'),
```
hassayag: Does `deep_get` do anything in these instances that couldn't just be done with `data.get('m_bDamagedByAbilities')`

LVL1024: True, `.get()` works fine here since it's just one level. I mostly used `deep_get` everywhere for consistency.

I can change it to `.get()` if you prefer it.

hassayag: It's a bit nitpicky, but yeah I think use the native .get where possible

LVL1024: Done

PR #264, opened 2025-11-24
fix: round numbers in ability upgrade descriptions

LVL1024:
Updated `string_utils.py` to:
1. Round variables to 3 significant figures.
2. Remove the `.0` from whole numbers.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/137) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/utils/string_utils.py
+++ b/src/utils/string_utils.py
@@ -62,12 +62,10 @@ def _replace_variables(desc, data):
             # strip out units of measure to prevent duplicates eg. "Cooldown reduced by 5ss"
             stripped_value = num_utils.remove_uom(value)
             if type(stripped_value) in [float, int]:
-                rounded_value = num_utils.round_sig_figs(stripped_value, 3)
+                if isinstance(stripped_value, float) and stripped_value.is_integer():
+                    return str(int(stripped_value))
 
-                if isinstance(rounded_value, float) and rounded_value.is_integer():
-                    return str(int(rounded_value))
-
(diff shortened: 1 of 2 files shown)


src/utils/string_utils.py
```diff
@@ -62,7 +62,12 @@ def replace_match(match):
             # strip out units of measure to prevent duplicates eg. "Cooldown reduced by 5ss"
             stripped_value = num_utils.remove_uom(value)
             if type(stripped_value) in [float, int]:
-                return str(stripped_value)
+                rounded_value = num_utils.round_sig_figs(stripped_value, 3)
```
hassayag: I think 4 or 5 significant figures here would be better, as its currently rounding too hard - eg. 103.5 -> 104

LVL1024:
> @LVL1024 Sorry for hijacking, I was testing out an idea and it worked in the end. The idea here is that rounding values can get a bit hairy, as we cover a wide range of values and want to avoid losing any precision. So instead I've opted to filter out "garbage" floats like 12.599999. Let me know what you think

@hassayag Oh yeah, that looks a lot better. And no worries about making changes, to be honest, if you ever see anything that you think could be done better in a PR I'm doing, please do so lol

hassayag (approved):

PR #271, opened 2025-12-12
feat: upgrade changelog system with hotfix support and smart edits

LVL1024:
**Key Changes:**
*   **Hotfix Detection:** Now groups forum posts by date (converted to Valve HQ time). Posts on the same day are merged; posts on subsequent days create new entries.
*  **Smart Appending:** Detects hotfixes by comparing fetched forum posts against the bot's local cache. If a new `=== Patch X ===` section is detected locally, it is extracted and appended to the bottom of the live Wiki page template, preserving any existing manual edits.
*   **Safety Limits:** The uploader is now restricted to the 10 most recent updates to ensure historical data is never accidentally modified.
*   **Navigation:** The bot now automatically populates the `| prev_update =` field using the `{{Update link}}` template.
*   **Chain Linking:** After uploading a new patch, the bot surgically edits the *previous* patch page to add the `| next_update =` link.
*   **Reply Linking:** Links to forum comments now use direct permalinks and append `(Reply)` to the source title for clarity.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/144) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/wiki/upload.py
+++ b/src/wiki/upload.py
@@ -123,7 +123,7 @@ class WikiUpload:
 
                     logger.info(f'Appending new hotfix section to Wiki page: {page_title}')
                     if not self.dry_run:
-                        page.save(new_page_text, summary='Deadbot: Appended new hotfix notes')
+                        page.save(new_page_text, summary=f'{self.upload_message}: Appended new hotfix notes')
                         logger.success(f'Successfully appended hotfix to {page_title}')
                 else:
                     logger.warning(f"Page {page_title} does not end with '}}', cannot safely append hotfix.")
@@ -248,7 +248,7 @@ class WikiUpload:
         logger.info(f'Linking {prev_page_title} -> {next_link_str}')
 
         if not self.dry_run:
-            page.save(new_text, summary=f"Deadbot: Linking next update to {latest_date.strftime('%Y-%m-%d')}")
+            page.save(new_text, summary=f"{self.upload_message}: Linking next update to {latest_date.strftime('%Y-%m-%d')}")
             logger.success(f'Updated {prev_page_title} with next link.')
 
     def _update_page(self, page, updated_text):


hassayag:
@LVL1024 Yet to have a deep dive on this one yet, but I have one question about the smart edits. 

The addition of updated patch notes with `Patch X` heading is a nice tool for editors. Is the intention that this new patch section is then manually combined into the rest of the page, with the `Patch X` heading removed? Or does this new `Patch X` heading have to stay there to stop the bot from reuploading it?

LVL1024:
> @LVL1024 Yet to have a deep dive on this one yet, but I have one question about the smart edits.
> 
> The addition of updated patch notes with `Patch X` heading is a nice tool for editors. Is the intention that this new patch section is then manually combined into the rest of the page, with the `Patch X` heading removed? Or does this new `Patch X` heading have to stay there to stop the bot from reuploading it?

@hassayag I was actually thinking the heading should stay. I think it's a nice way to organize the timeline of updates for that day, but technically it also needs to stay there so the bot knows the update is already on the page. If an editor deletes it, the bot will assume it's missing and re-upload it.

The heading is slightly different here, but I was thinking it'd be like this one from last year.
image

hassayag:
> @hassayag I was actually thinking the heading should stay. I think it's a nice way to organize the timeline of updates for that day, but technically it also needs to stay there so the bot knows the update is already on the page. If an editor deletes it, the bot will assume it's missing and re-upload it.

The bot dictating the end-user visuals of the wiki is something I would like to avoid. Maybe right now this is how we like it, but things can easily change in the future

There are two options I can see to avoid checking for this patch heading:
- Upon hotfix, manually delete the relevant update page and have the bot reupload it. Then copy parts of the previous version over to preserve manual edits - this should be okay if hotfixes aren't too frequent, albiet a little jank lol
- Use the `input-data` to compare to the last version instead of the wiki and add missing changelogs to the bottom accordingly. This would mean it only happens once from the bot's side, allowing editors to do as they please with it

Hopefully you get where I'm coming from, as its in our best interest to let editors choose how the pages should be structured

LVL1024:
@hassayag Good point, it definitely makes sense. I like option 2. Do you think it should add a new patch with a heading, or just post the changes only?

hassayag (approved):
This is great work mate, liking how the upload section is simplified now 🚀

PR #275, opened 2026-01-05
feat: upload all currently updated localizations to the wiki

LVL1024:
Added 22 more languages to the wiki upload list.

Added all available language files found in the game data (including those that are currently stale or empty) so the bot is ready for future updates.

New languages added:
* Lang pt-br.json (brazilian)
* Lang bg.json (bulgarian)
* Lang cs.json (czech)
* Lang da.json (danish)
* Lang nl.json (dutch)
* Lang fi.json (finnish)
* Lang fr.json (french)
* Lang de.json (german)
* Lang el.json (greek)
* Lang hu.json (hungarian)
* Lang id.json (indonesian)
* Lang it.json (italian)
* Lang ja.json (japanese)
* Lang es-419.json (latam)
* Lang no.json (norwegian)
* Lang pl.json (polish)
* Lang pt.json (portuguese)
* Lang ro.json (romanian)
* Lang sv.json (swedish)
* Lang th.json (thai)
* Lang uk.json (ukrainian)
* Lang vi.json (vietnamese)

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/146) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag:
> Should we stick to only uploading these active ones, or would it be better to add every available language file (including the stale/empty ones) so they are ready for the future?

I'd be happy with all of them being added, as the bot will only upload if the file is found on the wiki anyway

>What should the code for latam be? I used es-419 because it is the standard, but it is a bit weird. Should we just use latam instead?

If this happens we can change it to what people want (although seems unlikely we would have support for both spanish and latam realistically). We can use the standard es-419 code, as it has no effect on the wiki until the language file is manually created.

PR #284, opened 2026-01-23
refactor: overhaul NpcParser for automatic stat discovery

LVL1024:
Refactor `npc_units.py` from hardcoded parsers to a generic, recursive crawler.

### Key Changes:

* **Dynamic Discovery:** Automatically extracts all numeric/boolean stats from any NPC instead of using a hardcoded whitelist of supported units and attributes.
* **Audiovisual Filtering:** Strips cosmetic data (particles, sound events, models, decals) via a blocklist while preserving gameplay stats.
* **Automatic Unit Conversion:** Detects distance/velocity keywords and converts engine units to meters automatically.
* **Fixed Modifier Parsing:** Resolves bug where stats like `RespawnDelay` were lost when siblings to `m_vecScriptValues` (now extracts both).

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/151) - Reopen deadbot PR or run deploy workflow [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse_

### DIFF
--- a/src/parser/parsers/abilities.py
+++ b/src/parser/parsers/abilities.py
@@ -28,13 +28,13 @@ class AbilityParser:
         ability = self.abilities_data[ability_key]
 
         if type(ability) is not dict:
-            return None
+            return
 
         if 'm_eAbilityType' not in ability:
-            return None
+            return
 
         if ability['m_eAbilityType'] not in ['EAbilityType_Innate', 'EAbilityType_Signature', 'EAbilityType_Ultimate']:
-            return None
+            return
 
         ability_data = {
             'Key': ability_key,
@@ -127,7 +127,7 @@ class AbilityParser:
         elif 'm_strVAlue' in stat:
             value = stat['m_strVAlue']
         else:
-            return None
+            return
 
         return self._convert_stat(stat, key, value)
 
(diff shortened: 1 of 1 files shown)


LVL1024:
> Seems like a good change and makes things simpler to maintain. Do you know if this will have any issues on the wiki side eg. any changed keys?

Yes, this will introduce breaking changes that will require updates on the wiki side.

Because it will no longer use hardcoded keys or flatten objects, keys like `PlayerDamageResistance` are now `PlayerDamageResistPct`, and previously top-level stats like `WeaponRange` are now nested inside a `WeaponInfo` object.

The Lua module will need to be updated at the same time the change is pushed to master I think.

hassayag:
@LVL1024 
> The Lua module will need to be updated at the same time the change is pushed to master I think.

If it's something you're happy to fix on the fly we can merge, as I'm hoping to release all latest PRs tomorrow if possible. If not this can stay unmerged, up to you

LVL1024:
> If it's something you're happy to fix on the fly we can merge, as I'm hoping to release all latest PRs tomorrow if possible. If not this can stay unmerged, up to you

I can try to work on it tomorrow.

PR #303, opened 2026-02-09
feat: added parsing of item upgrades (enhanced item stats)

MonsterDomosed:
_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/167) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/parsers/items.py
+++ b/src/parser/parsers/items.py
@@ -131,6 +131,13 @@ class ItemParser:
                 parent_name = key
             self._add_children_to_tree(parent_name, parsed_item_data['Components'])
 
+        property_upgrades = self._parse_property_upgrades(item_value)
+        if property_upgrades:
+            parsed_item_data['PropertyUpgrades'] = property_upgrades
+
+        return parsed_item_data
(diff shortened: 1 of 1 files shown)


src/parser/parsers/items.py
```diff
@@ -131,6 +131,27 @@ def _parse_item(self, key):
...
+
+                if not prop_name or bonus_str is None:
+                    continue
+
+                try:
+                    bonus_value = num_utils.assert_number(bonus_str)
+                except (ValueError, TypeError):
+                    continue
+
+                property_upgrades.setdefault(prop_name, []).append(bonus_value)
```
hassayag: Shown in the output data, this is making each property something like
```
        "PropertyUpgrades": {
            "SlideScale": [
                50
            ],
```
Should it instead be
```
        "PropertyUpgrades": {
            "SlideScale": 50
```

PR #324, opened 2026-02-26
fix: remove duplicate LaunchWindowCooldown from Mirage Fire Scarabs

### LINKED ISSUE #308: Mirage Fire Scarabs has duplicate upgrade attribute
First upgrade shows "-15s Cooldown-15s Cooldown" due to there being two cooldown attributes that show the same thing. Only one should be stored in the ability cards output data

LVL1024:
Filters duplicate cooldown property from Mirage Fire Scarabs upgrade. 

Fixes #308 

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/183) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/parsers/ability_cards.py
+++ b/src/parser/parsers/ability_cards.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from utils import num_utils
 import utils.string_utils as string_utils
 from loguru import logger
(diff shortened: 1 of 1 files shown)


LVL1024:
> These kinds of edge case fixes are better suited for the wiki modules, but ideally we should resolve it in a generalised way.
> 
> One way we could do it is to only include upgrades that actually have a relevant property. These relevant properties include attributes in info sections or header attributes (stored in top-level of the ability card) - ie. AbilityCooldown, Radius etc.
> 
> In this case, `LaunchWindowCooldown` is none of these as it is kept in the `Cooldown` object, so by that logic, it should not be included in the upgrades list

Your approach would work but needs a hardcoded list. I used m_bCanSetTokenOverride instead, Valve sets this to true for canonical properties like AbilityCooldown, while internal ones like LaunchWindowCooldown don't have it. This achieves the same goal without hardcoding. Happy to switch if you prefer or if I overlooked a solution.

hassayag (approved):
Couldn't figure out what `m_bCanSetTokenOverride` meant, but this makes sense🚀

PR #334, opened 2026-03-07
feat: add Windows executable to releases

LVL1024:
Builds and releases `deadbot.exe` using PyInstaller on Windows runner. Tested via temporary workflow, executable runs correctly. 

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/190) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,6 +18,8 @@ env:
 jobs:
   release:
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
     permissions:
       contents: write
       packages: write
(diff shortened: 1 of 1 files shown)


hassayag:
So I've had a look into this, one main issue is the existing linux build includes a few things in the dockerfile, such as env vars and the depotdownloader executable, so the windows should ideally match that

In theory for any windows users, the current linux build is usable with Windows WSL enabled and using `wsl ./deadbot`. This should be included in the README however

And on top of that, I just tried to test the linux build with WSL and it has a bunch of issues that need resolving (which may also be a problem with a windows build)
- Subprocess logic does not work in the binary, such as when it tries to pull from the SteamDB repo
- The release action is uploading an ancient version, which I think is down to caching problems

I'm happy to deal with these issues as I made it the begin with, but for Windows I don't see enough demand to have native support when WSL is a really easy thing to enable

PR #335, opened 2026-03-09
feat: automate resource lookup data generation

LVL1024:
Currently, Module:Icon loads and iterates three separate JSON files on every page render, causing Lua timeouts on icon-heavy pages.

This PR adds ResourceLookupParser which pre-generates a single pre-indexed lookup table (ResourceLookup.json) at deploy time, allowing Module:Icon to do a direct lookup instead.

Only non-inferable fields are included in the data. Wiki-specific logic (image paths, links, CSS classes) is handled on the wiki side.

Module:Icon will need to be updated on the wiki when this is deployed..

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/191) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


src/parser/parsers/icon_lookup.py
```diff
@@ -0,0 +1,115 @@
...
+            existing = lookup.get(lower_name)
+
+            if existing is None:
+                lookup[lower_name] = {
+                    'name': name,
+                    'key': ability_key,
+                    'link': link,
+                    'image': name + '.png',
+                    'type': 'ability',
+                    'class': 'theme',
```
hassayag: What does "theme" refer to out of interest, is it the light/dark theming? 

In any case, could it be inferred on the wiki side. ie. if type is "ability", use specific class

hassayag: My point in the comment is to say that this `class` attribute doesn't seem necessary. The wiki can just apply this class as it wants for all `type=="ability` right?

PR #343, opened 2026-03-23
feat: added map entity info extraction and map image generator

ThePyrotechnic:
### The gist of it:

-  Add a new `download_files` method to `DepotDownloader` that can be used to download arbitrary files from the Deadlock depots
- Add a new `parse_map` flag (default `False`)
- Use that to selectively download the midtown VPK
  - Note that this requires Steam creds (and Deadlock access), same as with the localization stuff
- Create a new `GameMapParser` class
  - This relies on another new CLI param, `entity_helper_cmd`, which points to a [DeadlockEntityHelper](https://github.com/deadlock-wiki/DeadlockEntityHelper) binary
  - The new parser class feeds the downloaded map into the EntityHelper binary and uses the results to generate crate and statue plots
  - This adds a dependency on matplotlib
- Integrate the new parser into the `parser.py` workflow

Ultimately this results in a two outputs: 
- a new json output, called `midtown-metadata.json`. It's just the breakable counts for now.
- a new folder in the output, called `assets`. This is where I place the generated plots

### TODO

- Integrate this into the github workflow
- Get the resulting json and images onto the wiki

I tried to follow the existing conventions / control flow as best I could, but please do suggest changes.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/199) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag:
Great contribution! Just getting going with trying to run it
Downloaded the latest release of entity helper, and running the command returns this
```
wsl ../DeadlockEntityHelper extract --verbose ./game-data/game/citadel/maps/dl_midtown.vpk citadel_breakable_prop_wooden_crate initial_spawn_time_override double scales vector3 origin vector3

File does not exist: 'extract'.
Unrecognized command or argument 'citadel_breakable_prop_wooden_crate'.
Unrecognized command or argument 'initial_spawn_time_override'.
Unrecognized command or argument 'double'.
Unrecognized command or argument 'scales'.
Unrecognized command or argument 'vector3'.
Unrecognized command or argument 'origin'.
Unrecognized command or argument 'vector3'.

Description:
  Deadlock Entity Helper

Usage:
  DeadlockEntityHelper <vpk-file> <subclass> [options]

Arguments:
  <vpk-file>  The path to the midtown VPK file
  <subclass>  The entity subclass to filter for

Options:
  -v, --verbose   Enable verbose logging.
  --version       Show version information
  -?, -h, --help  Show help and usage information
```

Not sure if this is a quirk of windows WSL, but let me know if you can help resolve it

ThePyrotechnic:
@hassayag 
Nice - I uploaded an old binary to the release :D
Fixed: https://github.com/deadlock-wiki/DeadlockEntityHelper/releases/tag/v1.0.1
If you run `DeadlockEntityHelper --version` it should show `1.0.1+6ed27b9ce7e1cb27d1dd2f44a06b85830f3948ca`

hassayag (changes requested):
Managed to get it all working now, you'll see I've committed a change to use the process util and added some supprot for Windows WSL, as myself and other devs do (unfortunately) often use Windows

Other changes needed are any docker and github actions changes. Is that something you're able to handle? Happy to help if not

src/utils/process.py
```diff
@@ -11,23 +12,43 @@ def run_process(params, name=''):
...
-        with process.stdout:
-            for line in iter(process.stdout.readline, ''):
-                logger.trace(f'[process: {name}] {line.strip()}')
+        logger.trace(f'[process: {name}] Starting process with command {params}')
+        process = subprocess.Popen(params, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+
+        output_lines = []
+        for line in iter(process.stdout.readline, ''):
+            stripped = line.strip()
+            logger.trace(f'[process: {name}] {stripped}')
```
ThePyrotechnic: You are going to get many thousands of lines of output here, so we may not want to dump stdout - I wrote DeadlockEntityHelper to split any errors / trace info to stderr to avoid this. Most terminals will hit the scrollback limit which could be a pain for actual tracing

An alternative would be to change DeadlockEntityHelper to not pretty-print stdout by default (could add a flag for it)

.gitignore
```diff
@@ -14,4 +14,4 @@ docker-compose-local.yml
 .vscode
 exceptions.txt
 dist
-game-data
+poetry.toml
```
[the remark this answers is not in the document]
ThePyrotechnic: I like to put my virtual environments in the project folder, but poetry's default is a central cache directory. To override that I create a poetry.toml like this:
```
[virtualenvs]
in-project = true
```
This is meant for user-specific overrides according to their docs [here](https://deepwiki.com/python-poetry/poetry/2.3-configuration-management#local-vs-global-configuration):
> Note: Be careful about checking the poetry.toml file into your repository as it may contain user-specific settings.

src/utils/process.py
ThePyrotechnic: added `suppress_stdout` for this. Also added `stderr` output to the log

ThePyrotechnic:
I believe c36e69c covers the conversation so far. Let me know what you think
- combine localization and map download
- modify `depot_downloader.download_files` to support `file_list_path` and `files` simultaneously
- modify `process.run_process` to output stderr, optionally `suppress_stdout` in logs, use `with` to simplify closing stream handles
- clarify exceptions in GameMapParser
- use `--compact` output when calling DeadlockEntityHelper (requires v2.1.0)

ThePyrotechnic:
49a5965 should cover the docker container + GH Actions integration

src/steam/depot_downloader.py
```diff
@@ -28,19 +28,56 @@ def __init__(self, output_dir, deadlock_dir, depot_downloader_cmd, steam_usernam
         self.manifest_path = os.path.join(self.deadlock_dir, 'manifest.txt')
         self.force = force
```
hassayag: This force flag can be removed as its unused, seems depot downloader has its own caching that we can rely on instead

src/steam/depot_downloader.py
```diff
@@ -28,19 +28,56 @@ def __init__(self, output_dir, deadlock_dir, depot_downloader_cmd, steam_usernam
...
-            manifest_id = self._get_latest_manifest_id()
-
-        # Check if the manifest is already downloaded
-        downloaded_manifest_id = self._read_downloaded_manifest_id()
-        if downloaded_manifest_id == manifest_id and not self.force:
-            logger.info(f'Already downloaded manifest {manifest_id}')
-            return
-
-        self._download(manifest_id)
-        self._write_downloaded_manifest_id(manifest_id)
```
hassayag, line 43: can remove this function now as it is no longer used, check for any other unused functions

src/utils/parameters.py
```diff
@@ -60,6 +60,11 @@ def arg_group_base(parser):
         help='Forces decompilation even if game files and workdir versions match',
         default=is_truthy(os.getenv('FORCE', False)),
     )
+    group_base.add_argument(
+        '--entity_helper_cmd',
+        help='Path to DeadlockEntityHelper executable (also set with ENTITY_HELPER_CMD environment variable)',
+        default=os.getenv('ENTITY_HELPER_CMD', None),
```
hassayag, line 66: can add this to `.env.example` as well

src/utils/parameters.py
```diff
@@ -137,6 +142,12 @@ def arg_group_action(parser):
         help='Run the wiki upload in dry-run mode (also set with DRY_RUN environment variable)',
         default=is_truthy(os.getenv('DRY_RUN', False)),
     )
+    group_actions.add_argument(
+        '--parse_map',
+        action='store_true',
+        help='Parse map data from the game files (also set with PARSE_MAP environment variable)',
+        default=is_truthy(os.getenv('PARSE_MAP', False)),
```
hassayag, line 149: add to `.env.example`

hassayag (changes requested):
Great work, just a few minor bits

In terms of wiki integration, will the metadata and image files want to be uploaded on deployment?

ThePyrotechnic:
> In terms of wiki integration, will the metadata and image files want to be uploaded on deployment?

Yes, and I think I started that setup with this latest change (adding the json file to the map)
But I'm not sure how you want to handle the new `assets` folder?

Also cleaned up the `DepotDownloader` and updated the `.env.example`

hassayag:
> > In terms of wiki integration, will the metadata and image files want to be uploaded on deployment?
> 
> Yes, and I think I started that setup with this latest change (adding the json file to the map) But I'm not sure how you want to handle the new `assets` folder?
> 
> Also cleaned up the `DepotDownloader` and updated the `.env.example`

Yeah image upload will need a bit more work and can go in a separate issue and PR

hassayag (approved):
Let's ship it baby 🚀

Issue #346, opened 2026-04-06
Inconsistency for Items


Sophon:
Items have inconsistent values.

Some values are scaling values (so they have a scaling and a value) while some are string or a number.

For example, `cosmetic_item_snowball` has `"Damage": 1`. Meanwhile `upgrade_arctic_blast` has `"Damage" : { "Value": 175, "Scale":...`

I believe the most convenient solution would have all numeric values as scaling. Then the scaling can be 1 or whatever is needed.

Sophon:
@hassayag 
Alternative solution: have all numbers be raw numbers and then have nullable (so optional) scale value. 
Readability would not be affected.
File size in theory should stay the same - numbers that are scaled will have that property, raw numbers that don't simply won't.

hassayag:
@Sophon 
> numbers that are scaled will have that property, raw numbers that don't simply won't.

We already have scale values on items as far as I'm aware, implemented in the way you've described, but I might be misunderstanding you?

Sophon:
> [@Sophon](https://github.com/Sophon)
> 
> > numbers that are scaled will have that property, raw numbers that don't simply won't.
> 
> We already have scale values on items as far as I'm aware, implemented in the way you've described, but I might be misunderstanding you?

From the example, it's not the case.
What I meant was all would be a ScaledValue object but the scale, instead of being 1, would simply be absent.

hassayag:
> > [@Sophon](https://github.com/Sophon)
> > > numbers that are scaled will have that property, raw numbers that don't simply won't.
> > 
> > 
> > We already have scale values on items as far as I'm aware, implemented in the way you've described, but I might be misunderstanding you?
> 
> From the example, it's not the case. What I meant was all would be a ScaledValue object but the scale, instead of being 1, would simply be absent.

An example of how you would like the output data to look might help

Sophon:
Old:
```
"cosmetic_item_snowball": {
		"Name": "Snowball",
		"Cost": null,
		"Tier": null,
		"Activation": "InstantCast",
		"Slot": null,
		"Components": null,
		"TargetTypes": [
			"AllFriendly",
			"AllEnemy"
		],
		"ShopFilters": [
			"WeaponDamage"
		],
		"Damage": 1,
```

New:
```
"cosmetic_item_snowball": {
		"Name": "Snowball",
		"Cost": null,
		"Tier": null,
		"Activation": "InstantCast",
		"Slot": null,
		"Components": null,
[...]

PR #347, opened 2026-04-06
refactor (item): set value to scaled value instead of a raw number

Sophon:
For items, always use ScaledValue instead of sometimes using ScaledValue and sometimes just the raw number.

[Issue](https://github.com/deadlock-wiki/deadbot/issues/346)

### DIFF
--- a/src/parser/parsers/items.py
+++ b/src/parser/parsers/items.py
@@ -90,7 +90,10 @@ class ItemParser:
                 # Only filter if the value is a number and it is zero
                 if num_utils.is_zero(value):
                     continue  # Skip this zero-value attribute
-                parsed_item_data[attr_key] = value
+                parsed_item_data[attr_key] = {
+                    'Value': value,
+                    'Scale': None
+                }
             else:
                 logger.trace(f'Missing m_strValue attr in item {key} attribute {attr_key}')
 


hassayag:
@Sophon Since most values in the game do not have a scaler, we have favoured defaulting simply to a number for readability purposes, and an object for anything more complex, such as spirit scaling. Not sure I see a need to change that unless you could point one out?

Sophon:
> @Sophon Since most values in the game do not have a scaler, we have favoured defaulting simply to a number for readability purposes, and an object for anything more complex, such as spirit scaling. Not sure I see a need to change that unless you could point one out?

The problem is the inconsistency, since a number can be an object or a primitive. This guarantees the numbers are always an object and either have a scale at 1 or an acutally relevant number.

One example is in the issue: https://github.com/deadlock-wiki/deadbot/issues/346#issue-4214614717

hassayag:
> > @Sophon Since most values in the game do not have a scaler, we have favoured defaulting simply to a number for readability purposes, and an object for anything more complex, such as spirit scaling. Not sure I see a need to change that unless you could point one out?
> 
> The problem is the inconsistency, since a number can be an object or a primitive. This guarantees the numbers are always an object and either have a scale at 1 or an acutally relevant number.
> 
> One example is in the issue: [#346 (comment)](https://github.com/deadlock-wiki/deadbot/issues/346#issue-4214614717)

I understand what you mean, consistency is beneficial, but readability and file size are also bonuses with the current implementation (this change alone doubled the size of the item data output)

One thing to consider also is for consistency, if we changed it for items, we would do it for all data. That would then require a number of changes across the wiki's scripting and also increase . So I'm afraid I can't see a good enough reason to have to go through that effort

PR #358, opened 2026-05-18
feat: add IsImbued flag to item parser and item cards

LVL1024:
Adds an `IsImbued` boolean to identify items that can be imbued onto abilities, derived from `m_TargetAbilityEffectsToApply` in the raw item data.

- `items.py`: compute and include `IsImbued` in parsed item data
- `item_cards.py`: register `IsImbued` in `used_attributes` and the card dict

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/207) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/parsers/item_cards.py
+++ b/src/parser/parsers/item_cards.py
@@ -36,6 +36,7 @@ class ItemCardParser:
             'IsDisabled',
             'StreetBrawl',
             'PropertyUpgrades',
+            'IsImbued',
         ]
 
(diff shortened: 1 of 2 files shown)


LVL1024:
Closing because Domo's is better #362

Issue #361, opened 2026-05-19
Detect and add imbue metadata to ItemData from m_TargetAbilityEffectsToApply tags


MonsterDomosed:
The game files say if a item is a imbue in "m_TargetAbilityEffectsToApply" if they have the tag: 'CITADEL_TARGET_ABILITY_BEHAVIOR_IMBUE_ACTIVE', 'CITADEL_TARGET_ABILITY_BEHAVIOR_IMBUE_MODIFIER_VALUE' or 'CITADEL_TARGET_ABILITY_BEHAVIOR_IMBUE_ACTIVE_NON_ULT'

PR #362, opened 2026-05-19
feat: added imbue flag to item data

MonsterDomosed:
#361 

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/210) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/maps.py
+++ b/src/parser/maps.py
@@ -224,3 +224,14 @@ def get_section_type(value):
         raise Exception(f'{value} is not a valid item tooltip section type')
 
     return SECTION_TYPE_MAP[value]
+
+
+IMBUE_TAGS = {
+    'CITADEL_TARGET_ABILITY_BEHAVIOR_IMBUE_ACTIVE',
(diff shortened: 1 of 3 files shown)


LVL1024:
Hey, don't forget to update `item_cards.py` too, otherwise IsImbue will end up in the Other bucket in the output. Just needs to be added to used_attributes and the card dict, same pattern as IsDisabled and StreetBrawl.

hassayag (approved):

hassayag (changes requested):
Ah yeah just realised, as @LVL1024 mentioned, the output of item cards should place `IsImbue` on the top level, similar to the other flags

MonsterDomosed:
oops i forgot to remove them from other

hassayag (approved):
Good work 🚀

PR #367, opened 2026-05-23
feat: extract modifier hierarchy from abilities

Vergir:
Add parsing of all \*modifier\* fields in abilities. These fields are going to be exported and will include several hardcoded props + any int/float props. The hardcoded props like class/subclass will allow wiki maintainers to better use this in Lua automations to filter abilities that have some specific debuffs e.g. silence. The int/float props can be used to reference in abilities notes, like how melee buildup for Afterburn is a prop inside *modifier* field.

An example of how abilitydata.json would look like after this can be seen here
https://github.com/Vergir/deadlock-data/pull/1/changes

### DIFF
--- a/src/parser/parsers/abilities/__main__.py
+++ b/src/parser/parsers/abilities/__main__.py
@@ -3,6 +3,7 @@ from . import utils
 import utils.json_utils as json_utils
 import utils.num_utils as num_utils
 from .upgrades import parse_upgrades
+from .modifiers import parse_modifiers
 from loguru import logger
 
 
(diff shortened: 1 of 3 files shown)


hassayag (approved):
Looks good to me, nice first contribution 🫡 

I've moved your target branch to develop as that is the standard (unless we are hotfixing direct to master)

Vergir:
Thanks!

Is there anything else I should do, or do I just wait?

There is still
```
2 workflows awaiting approval
This workflow requires approval from a maintainer.
```
Looks like it needs some action from a maintainer?

hassayag:
> Thanks!
> 
> Is there anything else I should do, or do I just wait?
> 
> There is still
> 
> ```
> 2 workflows awaiting approval
> This workflow requires approval from a maintainer.
> ```
> 
> Looks like it needs some action from a maintainer?

Usually I let people merge their own PRs, but you may not have perms so I can do that

Normally we would run the github actions, but with the way things are layed out at the moment, only non-forked branches work with it (currently in the process of working out a more accessible solution)

I'll get you added to the organisation if you like, as that will be a nicer experience going forward

PR #370, opened 2026-05-24
Case insensitive key matching

Vergir:
Added a new `CaseInsensitiveDict` class to json_utils. It's a wrapper for dicts that allows matching keys and handles typos automatically. Used this new class wherever we load data for parsing. The goal is to protect against typos breaking deadbot in future patches.

Also removed all previous case-hacks since they're not needed now.

(As a small aside I've also changed call to copy version file so it's actually cross-platform and works on Windows now.)

### DIFF
--- a/src/decompiler/decompiler.py
+++ b/src/decompiler/decompiler.py
@@ -37,7 +37,7 @@ class Decompiler:
         self._import_localization_data()
 
         # Save version information to prevent unnecessary future decompiles
-        os.system(f'cp "{steam_inf_path}" "{version_path}"')
+        shutil.copy(steam_inf_path, version_path)
 
     def _import_game_data(self):
         """
--- a/src/parser/parser.py
+++ b/src/parser/parser.py
@@ -79,12 +79,16 @@ class Parser:
         # Convert .vdata_c to .vdata and .json
         scripts_path = 'scripts'
 
-        # Load json files to memory
+        # Load json files to memory. Each script tree is wrapped in a
+        # CaseInsensitiveDict so downstream parsers tolerate Valve's
+        # occasional capitalization typos (e.g. m_strVAlue, AbilitYCharges)
+        # without per-parser fallback code.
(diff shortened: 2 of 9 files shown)


src/decompiler/decompiler.py
```diff
@@ -37,7 +37,7 @@ def run(self):
         self._import_localization_data()
 
         # Save version information to prevent unnecessary future decompiles
-        os.system(f'cp "{steam_inf_path}" "{version_path}"')
+        shutil.copy(steam_inf_path, version_path)
```
Vergir, line 40: Cross-platform way to copy files, works on Windows now.

src/parser/parser.py
```diff
@@ -136,7 +137,6 @@ def _merge_localizations(self, language, group, data):
 
     def run(self):
         logger.trace('Parsing...')
-        os.system(f'cp "{self.DATA_DIR}/version.txt" "{self.OUTPUT_DIR}/version.txt"')
```
Vergir, line 139: this was unnecessary since the file is already copied in decompiler.py

src/utils/parameters.py
```diff
@@ -3,10 +3,13 @@
 from typing import Optional, Protocol
 from dotenv import load_dotenv
 
-from utils.string_utils import is_truthy
-
 load_dotenv()
 
+
+def is_truthy(value):
```
Vergir, line 9: moved this func here to avoid import cycle (this file was also the only usage)

src/utils/json_utils.py
```diff
@@ -4,6 +4,119 @@
 from utils import num_utils
 
 
+class CaseInsensitiveDict(dict):
```
Vergir, line 7: The meat of the PR.
Basically it keeps two dicts, one with raw keys and one with .lower()-ed keys and when fetching it tries to get from either.

Vergir:
moved here
https://github.com/deadlock-wiki/deadbot/pull/376/changes

PR #374, opened 2026-05-25
fix: defaulting to spirit scaling when scale type is missing

### LINKED ISSUE #326: Correct defaulting to spirit scaling in certain cases
Some cases default to spirit scaling when there is not an explicit `m_eSpecificStatScaleType` attribute, eg. Silver's bootkick

Instead we should use the `_class` attribute, such as `scale_function_tech_range` means it scales by `ETechRange`

LVL1024:
When m_eSpecificStatScaleType is missing, we were defaulting to spirit scaling. Now we infer the correct type from the _class attribute instead (e.g. scale_function_tech_duration → duration scaling).

Also filtered out 0.0 scale values on items to reduce output noise.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/216) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/maps.py
+++ b/src/parser/maps.py
@@ -201,6 +201,52 @@ SCALE_TYPE_MAP = {
 }
 
 
+def _class_to_scale_type(class_str: str) -> str | None:
+    """
+    Infer the human-readable scale type from a _class string.
+    Returns a value like 'spirit', 'range', 'duration', 'cooldown', etc.
+    """
+    if not class_str:
+        return None
+
+    # Pattern 1: scale_function_<suffix>
+    if class_str.startswith('scale_function_'):
+        suffix = class_str[len('scale_function_') :]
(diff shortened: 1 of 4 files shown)


src/parser/maps.py
```diff
@@ -201,6 +201,52 @@ def override_localization(attr):
 }
 
 
+def _class_to_scale_type(class_str: str) -> str | None:
```
hassayag: underscore before a function conventionally implies it is "private". This one is "public" as it is imported from the items parser

hassayag (changes requested):
Just some minor name change 🔥

src/parser/maps.py
LVL1024: Done

hassayag (approved):

PR #376, opened 2026-05-26
fix: added case insensitive key matching to account for game data typos

Vergir:
Added a new CaseInsensitiveDict class to json_utils. It's a wrapper for dicts that allows getting keys safely despite case typos. Used this new class wherever we load data for parsing. The goal is to protect against typos breaking deadbot in future patches. With that I also removed all previous case-hacks since they're not needed now.

(also some really small changes here and there, see comments)

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/218) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


src/decompiler/decompiler.py
```diff
@@ -37,7 +37,7 @@ def run(self):
         self._import_localization_data()
 
         # Save version information to prevent unnecessary future decompiles
-        os.system(f'cp "{steam_inf_path}" "{version_path}"')
+        shutil.copy(steam_inf_path, version_path)
```
Vergir, line 40: Cross-platform way to copy files, this makes it work on windows

src/parser/parser.py
```diff
@@ -137,7 +138,6 @@ def _merge_localizations(self, language, group, data):
 
     def run(self):
         logger.trace('Parsing...')
-        os.system(f'cp "{self.DATA_DIR}/version.txt" "{self.OUTPUT_DIR}/version.txt"')
```
Vergir, line 140: this was unnecessary since the file is already copied in decompiler.py

src/utils/parameters.py
```diff
@@ -3,10 +3,13 @@
 from typing import Optional, Protocol
 from dotenv import load_dotenv
 
-from utils.string_utils import is_truthy
-
 load_dotenv()
 
+
+def is_truthy(value):
```
Vergir: moved this func from string_utils to avoid import cycle (this file was also the only usage)

src/utils/json_utils.py
```diff
@@ -4,6 +4,133 @@
 from utils import num_utils
 
 
+class CaseInsensitiveDict(dict):
```
Vergir, line 7: The meat of the PR.
Basically it keeps two dicts, one with raw keys and one with .lower()-ed keys and when fetching it tries to get from either.

Vergir:
@hassayag gentle ping :)

src/utils/parameters.py
hassayag: I like this in the utils since they are generic enough to have use elsewhere in future. What's the import cycle issue exactly?

hassayag (commented):
Great idea, it's even fixed a couple typos I had never noticed. All looks good, just one question about whether we need to move that util

src/utils/parameters.py
Vergir: The cycle
```
parameters.py   ──► from utils.string_utils import is_truthy
                      ↓
string_utils.py  ──► from utils import num_utils
                      ↓
num_utils.py     ──► from constants import ENGINE_UNITS_PER_METER
                      ↓
constants.py     ──► from utils import parameters
                      ↓
(back to the top) parameters.py ...
```

You can see it on current develop branch, it exists but it doesn't break anything because of how Python resolves import order. In current entry point (deadbot.py), things partially resolve before each circular edge is followed, so it works okay. However any module that imports utils.string_utils by itself (e.g. some unit test in future) crashes with ImportError. That's what happened when I was working on this PR.

To be fair, my fix to move `is_truthy` may have a been rushed. Looking at the cycle above I think a better way to break it is in `constants.py import parameters.py` step. I think constants shouldn't import code modules and execute it, so I left args parsing in parameters.py and changed deadbot.py from `args = constants.ARGS` to `args = parameters.ARGS`. And is_truthy is now back in string_utils.py.

See if you like it more?

hassayag: Yeah agreed, not a clue why ARGS was being exported from there

hassayag (approved):
Ship it 🚀 🔥

PR #383, opened 2026-05-29
feat: extract gravity/air control stats to hero level

LVL1024:
Some heroes (currently Celeste and Ivy) have gravity and air control stats defined on their weapon ability instead of in hero data. This extracts those and promotes them to top-level hero stats so they're easier to find and query.

Celeste gets AirControlPercent, AirControlAccelPercent, and GravityChange. Ivy gets GravityChange.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/225) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag (approved):
💯 💯

PR #384, opened 2026-05-31
feat: export statemasks to modifiers

Vergir:
This is extending our Ability Modifiers extractions with StateMask data.

Exporting modifiers allowed us to automate list of affected abilities on
https://deadlock.wiki/Dispel_Magic

Now I want to export more data: StateMasks that tell what states are applied (stun, immobilize, etc) with buffs/debuffs.
This also removes exporting of `m_vecAutoRegisterModifierValueFromAbilityPropertyName`: this field provides more harm than good and is not used anywhere anymore.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/226) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


Vergir:
Not sure why it sikpped parsing, but anyway here's how it will look
https://github.com/deadlock-wiki/deadlock-data/pull/228/changes

hassayag:
Does this cause any breaking changes on the wiki? Such as a module that uses the old class names

Vergir:
> Does this cause any breaking changes on the wiki? Such as a module that uses the old class names

Not at all. Conversely, it is specifically ignored in one module's code so we'll be able to make the module cleaner by merging this :)

hassayag (approved):

PR #391, opened 2026-06-08
ci: fix race condition causing deploy to skip on new PRs

### LINKED ISSUE #389: Fix parser not being run when PR is initially opened
Due to deploy and pr-open script being run simultaneously, the deploy script skips parsing as the deadlock-data branch is not yet created. Instead we should run deploy via workflow-call at the end of pr-open to mitigate this race condition

LVL1024:
Fixes deploy workflow racing ahead of pr-open and skipping because the deadlock-data branch doesn't exist yet.

- `pr-open.yaml` now triggers deploy after branch creation
- Removed `pull_request` trigger from `deploy.yaml`
- Added path filter to preserve existing deploy behavior

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/239) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag (approved):
Good work, not sure why I put this off for so long.. I just modified the path-filter as they apparently work a little differently. If that looks good to you then feel free to merge

PR #394, opened 2026-06-16
feat: parse ability behaviour bits into BehaviourBits field

Vergir:
Added parsing for AbilityBehaviourBits. This field is not parsed in any way right now, and adding it to AbilityData.json will allow us to make automated list of abilities that get disabled by https://deadlock.wiki/Slowing_Hex and similar effects (Whaith's ult etc).

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/237) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag (approved):
Noice 🚀

PR #397, opened 2026-06-22
feat: parse Street Brawl item draft bucketing

Vergir:
Parse m_mapItemDraftBucketing and m_mapItemDraftCounterWeights into a heradata.json. This will allow to list items that are more/less likely to be drafted for each hero in Street Brawl.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/241) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


Vergir:
One thing I'm a bit on the fence about is that it our `hero_data` gets quite bigger: from 200kb to 1.6mb. Lines count from 6k to 52k. CSV is roughtly same. We can move the new data to a separate file, like street_brawl.json, but I'm not sure if it's a good idea, to separate it like this. In the end i chose to stay in hero_data, but open to any corrections here.

hassayag:
> One thing I'm a bit on the fence about is that it our `hero_data` gets quite bigger: from 200kb to 1.6mb. Lines count from 6k to 52k. CSV is roughtly same. We can move the new data to a separate file, like street_brawl.json, but I'm not sure if it's a good idea, to separate it like this. In the end i chose to stay in hero_data, but open to any corrections here.

Yeah I'd agree it is too large for the hero json file, especially if we think this would be used on a page separate from each hero's main page. Have you seen any patterns in the data that show certain heroes share X set of weights? As I can see that one of the sets is used across 24 heroes

Vergir:
> Yeah I'd agree it is too large for the hero json file, especially if we think this would be used on a page separate from each hero's main page. Have you seen any patterns in the data that show certain heroes share X set of weights? As I can see that one of the sets is used across 24 heroes

The 24 heroes you mentioned is a set used for unreleased heroes (Raven-Slork etc). I also found that there's a default set of 178 items that are used for all heroes that can be minimized.

I've done some data normalization, and it's much smaller now: from 1.2mb to 200kb. Also I redirected data to a separate file `street-brawl-data.json`. See if you like it better now?

Vergir:
I've redone the parsing with the latest changes, see here
https://github.com/deadlock-wiki/deadlock-data/pull/241/changes

Also, if any comments or further edits are required, I suggest to move this discussion to discord server, it's way easier to communicate there :)

hassayag (approved):
Looks great, the default/outlier system is a nice way of laying it out 🚀

PR #401, opened 2026-07-03
feat: migrate changelog fetching from Deadlock forums to Steam API

### LINKED ISSUE #355: Fix changelogs
The devs have started posting some patch notes to Steam, but they still sometimes post them on the forums. At the same time, they also sometimes post a link to the Steam patch notes on the forums, which makes the bot upload the Steam link as a patch note.

LVL1024:
Here is the updated PR description in a markdown codeblock, ready to copy and paste:

Changed from using the forums to the Steam API since the devs don't post patch notes on the forums anymore, they moved to Steam posts. Also removed hero labs code.

What changed:
- `fetch_forum_changelogs()` → `fetch_steam_changelogs()`, reads from `steam_community_announcements` filtered to `patchnotes`-tagged posts
- `_bbcode_to_text()` converts Steam BBCode to plain text (paragraphs, lists, links, images, embeds, plus a catch-all for unknown tags)
- MD5 hashing (`steam_hash`) detects when Valve edits a post after publishing. If it changes, the local `.txt` file is updated silently, and hotfix detection is adjusted to prevent false positives.
- `STEAM_MIGRATION_DATE` cut-off (2026-06-04) keeps all historical forum-sourced data intact
- `forum_id` → `source_id` in `changelog_configs.json`; `title`, `steam_hash` fields added
- `feedparser` and `bs4` dropped from `pyproject.toml`; `tzdata` added
- Removed Hero Lab changelog parsing: The `get_gamefile_changelogs()` pipeline (~130 lines) parsed `citadel_patch_notes_english.json` into `_HeroLab` entries. Devs haven't used Hero Labs in forever.
- Rewrote the changelog wikitext formatter to group entries under `== Items ==` and `== Heroes ==` sections (with `=== Hero/Item ===` and `==== Ability ====` sub-headings) instead of flat bullet points, matching the wiki's Patch Formatter output.
- Added `hide_general` to `parse_changelogs.py` to suppress the General heading on patches with no unclassified bullets.
- Added `Unstable Rift` and `Rift Trooper` to `link_targets.json`.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/244) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


src/changelogs/parse_changelogs.py
```diff
@@ -129,10 +126,18 @@ def format_and_save_wikitext_changelogs(self, changelogs, changelog_configs):
...
-            if source_link:
-                # Logic to extract title from URL
+            source_link = config.get('link') or ''
+            source_title = config.get('title')
+
+            is_steam_link = (
+                'store.steampowered.com/news/' in source_link
+                or 'steamstore-a.akamaihd.net/news/' in source_link
+                or 'steamcommunity.com/app/' in source_link
+            )
```
hassayag: Is there any circumstance that the link would not be steam if we are pointing to the steam rss feed?

src/changelogs/fetch_changelogs.py
```diff
@@ -197,257 +219,89 @@ def changelogs_to_file(self):
...
+                        if existing_config.get('steam_hash') and existing_config['steam_hash'] != e['steam_hash']:
+                            logger.warning(f'Steam post for {date_key} was edited by the devs!')
+                            self.changelog_configs[changelog_id] = {
+                                'source_id': e['version'],
+                                'date': date_key,
+                                'link': e['link'],
+                                'is_hero_lab': False,
+                                'title': e.get('title'),
+                                'steam_hash': e['steam_hash'],
+                                'was_edited': True,
```
hassayag: Interested to know how we would handle edited updates. Would it function with our current hotfix system?

src/changelogs/fetch_changelogs.py
```diff
@@ -456,47 +310,36 @@ def fetch_forum_changelogs(self):
...
 
-                # Skip if this content is already in final_text
                 normalized_entry = re.sub(r'^- ', '* ', entry_text, flags=re.MULTILINE)
                 normalized_final = re.sub(r'^- ', '* ', final_text, flags=re.MULTILINE)
                 if normalized_entry not in normalized_final:
                     patch_num = get_next_patch_num(final_text)
                     final_text += f'\n\n=== Patch {patch_num} ===\n\n{entry_text}'
                     logger.debug(f'Adding Patch {patch_num} to {date_key}')
 
-            # Detect hotfixes - compare new patches against old content
```
hassayag, line 499: Please be observant what your AI is spitting out, they often strip comments

src/changelogs/fetch_changelogs.py
```diff
@@ -505,102 +348,34 @@ def get_next_patch_num(txt):
...
+                source_id = main_entry['version']
                 link = main_entry['link']
+                title = main_entry.get('title')
+                steam_hash = main_entry.get('steam_hash')
             elif existing_config:
+                source_id = existing_config.get('source_id')
                 link = existing_config.get('link')
+                title = existing_config.get('title')
+                steam_hash = existing_config.get('steam_hash')
             elif append_entries:
```
hassayag: Correct me if I'm wrong as I'm not that familiar with the changelog side

Is `append_entries` a remnant of the old forums, which had a main forum post with additional updates added as comments? If so, have you seen a similar thing with their use of steam news?

src/changelogs/fetch_changelogs.py
```diff
@@ -505,102 +348,34 @@ def get_next_patch_num(txt):
...
+                source_id = append_entries[0]['version']
                 link = append_entries[0]['link']
+                title = append_entries[0].get('title')
+                steam_hash = append_entries[0].get('steam_hash')
```
hassayag: (if possible,) I would try to make use an object like `{source_id, link, title, steam_hash}`, although you would have to deal with the `version` vs `source_id` issue

hassayag (commented):
Sorry I haven't gotten to this sooner. Great to see this has worked nicely. Main thing I am concerned with is getting rid of any old code that was designed specifically for the old RSS feed

src/changelogs/parse_changelogs.py
LVL1024: Any changelog before 2026-06-04 used the old forum links. But to keep things clean, I just made the parser skip generating wikitext for anything before that migration date entirely.

src/changelogs/fetch_changelogs.py
LVL1024: Fixed, will check it better in the future.

src/changelogs/fetch_changelogs.py
LVL1024: It was originally for forum comments. But we still need it for Steam in case Valve posts multiple patch notes on the same day. Since the wiki only has one page per date, we still use `append_entries` to merge them under `=== Patch N ===` headers.

src/changelogs/fetch_changelogs.py
LVL1024: I removed the manual `was_edited` flag entirely. Now, if Valve edits a post, the bot detects the hash change and silently updates the local .txt file. To prevent false hotfixes when overwriting, I set it so the newly fetched text is treated as the "old text" for comparison. The hotfix system still runs fine, and the wiki stays safe because the uploader skips overwriting the main page (while still appending genuinely new hotfix sections via the separate appender).

src/changelogs/fetch_changelogs.py
LVL1024: I collapsed that whole block down by determining the source dict first: `source_data = main_entry or existing_config or (append_entries[0] if append_entries else {})`. It should handle the version vs source_id key discrepancy.

LVL1024:
I also wrapped the Steam API parsing in error handling to prevent malformed news items from crashing the changelog run. Also removed some leftover forum and hero lab code.

LVL1024:
Other than fixing any other issues you might spot, we'll have to change the output of the wiki patch uploads to match closer to the formatting used there, as it has been changed.

hassayag:
Looks great to me. Make any further changes and set it to ready for review, and I'll give it a final look

hassayag (dismissed):
Seems good to me, but part of that is pure faith 😂

PR #405, opened 2026-07-07
feat: upload map image assets to the wiki

### LINKED ISSUE #348: Upload map image assets to the wiki
New image assets added by #343 are not currently handled by the upload process. 

- Add new image files to the `pages.py` file list
- Add support for uploading those images to the `File` namespace

LVL1024:
Uploads the three map plot images (golden statues, crate locations, shop locations) generated by the map parser to the wiki File namespace.

Adds a `--force-asset-upload` flag that can be toggled via workflow dispatch to re-upload existing images when map data changes.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/246) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/wiki/upload.py
+++ b/src/wiki/upload.py
@@ -5,7 +5,7 @@ import re
 from datetime import datetime
 from typing import List, Tuple
 from utils import json_utils, game_utils, meta_utils
-from .pages import DATA_PAGE_FILE_MAP, IGNORE_PAGES
+from .pages import DATA_PAGE_FILE_MAP, IGNORE_PAGES, IMAGE_FILE_MAP
 from loguru import logger
 from . import changelog_utils
 
@@ -49,6 +49,7 @@ class WikiUpload:
     def run(self):
         logger.info(f'Uploading Data to Wiki - {self.upload_message}')
         self._update_data_pages()
+        self._upload_assets()
         self.wiki_updates = self._get_existing_update_pages()
         self._upload_changelog_pages()
         self._process_hotfixes()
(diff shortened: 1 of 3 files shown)


src/wiki/upload.py
```diff
@@ -214,6 +219,41 @@ def _update_data_pages(self):
...
+            # Check if the file already exists on the wiki to avoid
+            # creating unnecessary file revisions. Unlike page saves,
+            # MediaWiki always creates a new revision on file upload,
+            # even if the binary content is identical.
+            # Use --force-asset-upload to override and re-upload existing files.
+            page_title = f'File:{wiki_filename}'
+            page = self.site.pages[page_title]
+            if page.exists and not self.force_asset_upload:
+                logger.trace(f'File "{page_title}" already exists on wiki, skipping.')
+                continue
```
hassayag: Files need to be handled automatically, can't rely on us checking a box. You mention that file upload always creates a new revision, but I had a play on the wiki itself and it blocks it. Have you tested that via the python client?

LVL1024: I couldn't test the live API earlier due to Codespace/Cloudflare blocks, which is why I left the manual guard in initially.
I've now removed the `--force-asset-upload` flag and the `page.exists` check entirely, and retested the updated code.

hassayag (approved):

PR #409, opened 2026-07-20
ci: add workflow to automate release PR creation

LVL1024:
Adds a `create-release-pr.yaml` workflow that can be triggered manually via `workflow_dispatch` with a version bump type input (patch/minor/major). It checks out the develop branch, computes the next version, creates a `release/vX.X.X` branch, pushes it, and opens a pull request against `master`.

### DIFF
--- a/.github/workflows/create-release-pr.yaml
+++ b/.github/workflows/create-release-pr.yaml
@@ -1,0 +1,66 @@
+name: Create Release PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: 'Type of version bump'
+        required: true
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
+        default: patch
+
+jobs:
+  create-release-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout develop branch
+        uses: actions/checkout@v4
+        with:
+          ref: develop
+          fetch-depth: 0
+          token: ${{ secrets.GH_TOKEN }}
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Configure Git
+        run: |
+          git config user.name "Deadbot0"
+          git config user.email "deadbot1101@gmail.com"
+
+      - name: Compute next version
+        id: version
+        run: |
+          git fetch origin master
+          # Extract the clean MAJOR.MINOR.PATCH from master, ignoring develop's beta suffix
+          MASTER_VERSION=$(git show origin/master:pyproject.toml | grep -Po '(?<=version = ")[0-9]+\.[0-9]+\.[0-9]+')
+          
+          # Calculate the next version using our refactored script
+          NEXT_VERSION=$(python3 ./scripts/increment_version.py ${{ inputs.bump }} --base $MASTER_VERSION --print-only)
+          
+          echo "Next version: $NEXT_VERSION"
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release-pr.yaml
```diff
@@ -0,0 +1,43 @@
+name: Create Release PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version number for the release (e.g., 1.15.0)'
+        required: true
+        type: string
```
hassayag: Would be better as an input of "major" "minor" or "patch", as that removes even more thinking from the equation. See https://github.com/deadlock-wiki/deadbot/blob/e410ddf0bd85ecabfc0592862fde14b4c399c0bb/.github/workflows/increment-version-dev.yaml#L27-L28 which uses an existing script to bump "beta"

hassayag: (might need to modify or make a copy of the function to allow just getting the next version number, rather than actually bumping it)

LVL1024: Is what I have now more what you were looking for?

LVL1024:
Also, I cant figure out why the lint check is failing.

Edit: Figured it out. pre-commit uses a pinned ruff version (0.6.2) while the CI uses the latest version. And the latest version of ruff came out yesterday and has breaking changes.

https://github.com/astral-sh/ruff/releases

.github/workflows/create-release-pr.yaml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          git config user.name "Deadbot0"
+          git config user.email "deadbot1101@gmail.com"
+
+      - name: Compute next version
+        id: version
+        run: |
+          git fetch origin master
+          # Extract the clean MAJOR.MINOR.PATCH from master, ignoring develop's beta suffix
+          MASTER_VERSION=$(git show origin/master:pyproject.toml | grep -Po '(?<=version = ")[0-9]+\.[0-9]+\.[0-9]+')
```
hassayag: Is this necessary if the script already reads the version for you?

.github/workflows/create-release-pr.yaml
hassayag: Yeah looks great, just one comment to deal with

hassayag:
> Also, I cant figure out why the lint check is failing.
> 
> Edit: Figured it out. pre-commit uses a pinned ruff version (0.6.2) while the CI uses the latest version. And the latest version of ruff came out yesterday and has breaking changes.
> 
> https://github.com/astral-sh/ruff/releases

Ah interesting, I'll sort that out in a separate PR

.github/workflows/create-release-pr.yaml
LVL1024: Yes, it's necessary. The workflow checks out `develop`, which has a beta-suffixed version (e.g., `1.16.0-beta.4`). The script's `read_version()` reads from the current `pyproject.toml`, so without `--base` the bump math would be calculated against that beta version rather than the last stable release on `master`. Fetching `master` and passing it via `--base` is exactly why that flag was added.

hassayag: The script is designed to handle beta versions. Eg. 1.16.0-beta.4 minor release would return 1.17.0

LVL1024: Ah, my mistake. I misunderstood the script on that part.

hassayag (approved):
Noice, will test on the next release 🤞

PR #413, opened 2026-07-26
feat: added street brawl ability changes

Vergir:
Add code to street_brawl.py that re-parses ability-data and extracts only Street Brawl specific changes, then writes it to street-brawl-data.json.

We have ability changes on the Street Brawl Page
https://deadlock.wiki/Street_Brawl#Heroes
But this list is outdated due to being 100% manual.

The re-parsing is done in a way that mirrors object structure from ability-data.json so it can be parsed by wiki modules in the same way.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/254) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag (approved):
The street brawl fans will be happy with this 🙌

PR #430, opened 2026-07-27
feat: added scaling data for Duration/Range/Charges

Vergir:
Added parsing scaling for ability range, duration, radius, and some other types which were ignored before because they're living in another field.

Example changes can be seen in the attached deadbot/data PR.

---

This introduces a couple of risky changes to ability-data.json and ability-cards.json
- Some fields go from being flat `"AbilityDuration": 25` to nested `"AbilityDuration": {"Value": ..., "Scale": ...}`.
- Scalings can now be arrays. For example, Doorman's Doorway scales with both SP and Range. Previously Range was ignored, now it is included.

I've already tested the changes by hand patching changes onto live wiki and fixing modules that were not prepared for that, but wanted to note anyway.

---

Another note, this PR does not touch items (which also often scale with ability range and duration), as I didn't want to increase risk further. Items can get a similar addition in the future.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/258) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_

### DIFF
--- a/src/parser/maps.py
+++ b/src/parser/maps.py
@@ -1,3 +1,5 @@
+from loguru import logger
+
 TARGET_TYPE_MAP = {
     'CITADEL_UNIT_TARGET_ALL_ENEMY': 'AllEnemy',
     'CITADEL_UNIT_TARGET_ALL_FRIENDLY': 'AllFriendly',
@@ -235,28 +237,6 @@ SCALE_TYPE_MAP = {
     'EWeaponPower': 'weapon_power',
 }
 
-# Scale types the game files use that are of no interest to the wiki, either because they only
-# apply to data we do not export, such as items and weapon sets, or because they say nothing
-# beyond what the attribute itself already does. Listing them keeps get_scale_type free to
-# raise on a scale type that is genuinely new to us
-IGNORED_SCALE_TYPES = {
-    'EAirMoveDistanceScale',
-    'EBuildUpRate',
-    'EChannelDuration',
-    'EClipSizeIncrease',
-    'EDamageScale',
-    'EItemCooldown',
-    'EMeleeRange',
-    'EProcBuildUpRateScale',
-    'EReloadSpeed',
-    'ETechDamageScale',
-    'EWeaponFalloffMaxRange',
-}
-
-
-def is_ignored_scale_type(scale):
-    return scale in IGNORED_SCALE_TYPES
-
 
 def class_to_scale_type(class_str: str) -> str | None:
     """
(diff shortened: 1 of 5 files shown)


hassayag:
@Vergir The concept of scalings of cooldowns on cooldown, ranges on range etc is a bit confusing. Mostly also seem to be 1's as well, so feels redundant

Vergir:
> @Vergir The concept of scalings of cooldowns on cooldown, ranges on range etc is a bit confusing. Mostly also seem to be 1's as well, so feels redundant

Because these scalings live in many props. Consider an example for Ability Range:
- Paige's dragon scales BOTH its travel range and its radius
- Dynamo's Pulse scales it's travel range but NOT its radius
- Doorman's Call Bell scales it's radius but NOT it's travel range.

Right now this info is missing altogether, but can be revealed if we parse and write scaling info for the related fields. I added parsing for ALL fields that have these. Of course, besides above mentioned props most `AbilityCastRange` got them 'cause most ranges scale.

If you're worried about the size of the file we can filter some of these out. Though we'd lose understanding if an ability doesn't scale something because it was filtered or because it actually doesn't scale: Chain Gang is an example of `AbilityDuration` that intentionally doesn't have scaling. So I'd suggest to keep current PR as is and instead cut some actually useless things, like completely unused `AbilityUnitTargetLimit` or  `AbilityCooldownBetweenCharge` with a value of -1 for all abilities except a handful of charged ones. We also have ~9 records for Graves' abilities, mostly scrapped during development.

hassayag:
> > @Vergir The concept of scalings of cooldowns on cooldown, ranges on range etc is a bit confusing. Mostly also seem to be 1's as well, so feels redundant
> 
> Because these scalings live in many props. Consider an example for Ability Range:
> 
>     * Paige's dragon scales BOTH its travel range and its radius
> 
>     * Dynamo's Pulse scales it's travel range but NOT its radius
> 
>     * Doorman's Call Bell scales it's radius but NOT it's travel range.
> 
> 
> Right now this info is missing altogether, but can be revealed if we parse and write scaling info for the related fields. I added parsing for ALL fields that have these. Of course, besides above mentioned props most `AbilityCastRange` got them 'cause most ranges scale.
> 
> If you're worried about the size of the file we can filter some of these out. Though we'd lose understanding if an ability doesn't scale something because it was filtered or because it actually doesn't scale: Chain Gang is an example of `AbilityDuration` that intentionally doesn't have scaling. So I'd suggest to keep current PR as is and instead cut some actually useless things, like completely unused `AbilityUnitTargetLimit` or `AbilityCooldownBetweenCharge` with a value of -1 for all abilities except a handful of charged ones. We also have ~9 records for Graves' abilities, mostly scrapped during development.

Okay you do make sense, while we all know that AbilityCooldown changes with cooldown reduction, it is nice to have that in writing, especially if we have edge cases. 

I don't mind `-1`, wouldn't have a major impact on file size anyway, and could break things on the wiki as I think it expects it in some places

One thing I will mention is that I am trying to keep ability-data and ability-cards in sync, so any raw data (not related to card UI), should be handled by ability-data parser, and then simply passed over to ability cards

Vergir:
> One thing I will mention is that I am trying to keep ability-data and ability-cards in sync, so any raw data (not related to card UI), should be handled by ability-data parser, and then simply passed over to ability cards

Glad you brought that up, I was thinking about it.
I made it so the scalings are parsed once and then pass down to both ability-data and abililty-cards, which sounds like how you see it too. On wiki side the new scalings won't be drawn because module already knows about them and hides them. But if I misunderstood and you'd rather have these scalings be only on ability-data and not on ability-cards, say so and I'll make the change.

hassayag:
> Glad you brought that up, I was thinking about it.
> I made it so the scalings are parsed once and then pass down to both ability-data and abililty-cards, which sounds like how you see it too. On wiki side the new scalings won't be drawn because module already knows about them and hides them. But if I misunderstood and you'd rather have these scalings be only on ability-data and not on ability-cards, say so and I'll make the change.

Okay ignore me, I misread the diff, looks like its in both ability-data and cards, so you're all good actually

hassayag (approved):

PR #437, opened 2026-08-02
feat: omit scalings that apply at a rate of zero

Vergir:
Don't write scaling to ability-data / ability-cards if it is zero.

Since `scaling: 0` is the same as no scaling (I've checked all abilities to make sure it's true) we can just omit it to say the same thing in less space, and it means an editor reading ability-cards.json can know a scaling is present when they see it, rather than checking its value before believing it

Also items.py has left them out of item data all along, so this brings the two parsers into line.

Also it would omit weird zero scalings from ability cards for [Doorway](https://deadlock.wiki/The_Doorman#Doorway) and [Fixation](https://deadlock.wiki/Haze#Fixation) and make them look closer to their in-game counterparts.

_Parsed data in [deadlock-data PR](https://github.com/deadlock-wiki/deadlock-data/pull/263) - Reopen deadbot PR or run deploy workflow for this branch [here](https://github.com/deadlock-wiki/deadbot/actions/workflows/deploy.yaml) to reparse the data_


hassayag:
@Vergir I haven't checked every single one, but these 0 values are paired with a corresponding upgrade. For the wiki upgrade gadget to work, it needs an outline of where those values are, but hides them when they are zero. So potentially the fix to those specific abilities is to fix the bug on the wiki side?

Vergir:
> @Vergir I haven't checked every single one, but these 0 values are paired with a corresponding upgrade. For the wiki upgrade gadget to work, it needs an outline of where those values are, but hides them when they are zero. So potentially the fix to those specific abilities is to fix the bug on the wiki side?

Upgrades is one thing and it can be fixed on the wiki side only, but there's also a benefit of just making data clearer: some zero scalings get replaced by upgrades, but many zeroes there are there just because it is there in files, and it adds nothing but confusion. Then there's argument of unifying the behaviour for abilities and items.

I'm not too attached to it so if you want these to stay we can just decline this, I just thought it would be better with the proposed behaviour

hassayag:
> > @Vergir I haven't checked every single one, but these 0 values are paired with a corresponding upgrade. For the wiki upgrade gadget to work, it needs an outline of where those values are, but hides them when they are zero. So potentially the fix to those specific abilities is to fix the bug on the wiki side?
> 
> Upgrades is one thing and it can be fixed on the wiki side only, but there's also a benefit of just making data clearer: some zero scalings get replaced by upgrades, but many zeroes there are there just because it is there in files, and it adds nothing but confusion. Then there's argument of unifying the behaviour for abilities and items.
> 
> I'm not too attached to it so if you want these to stay we can just decline this, I just thought it would be better with the proposed behaviour

I can see it being cleaner purely from a data perspective. But since the lua/js on the wiki would need a bunch of changes, I'd rather minimise effort/risk until we have better reason to do it

I can't be sure, but it's possible we could keep those zero values on ability-cards.json, but not the ability-data.json. Since ability-cards is designed for the wiki idm that staying a bit messier. But would need to verify if that would work

Vergir:
Lua modules won't need changes, I tested it. Upgrade gadget might require update and I'm willing to do it, but if we would rather not risk it then it seems better to just decline PR

